PR #1, opened 2017-06-02
Add the example as an example test program.

sunfishcode:
Putting it in the examples directory makes it be automatically built as
part of `cargo test`.


yurydelendik:
Thank you for the patch

PR #10, opened 2017-06-03
Remove support for concatenated modules.

sunfishcode:
The spec testsuite now tests that concatenated modules are rejected:
    
    https://github.com/WebAssembly/testsuite/blob/master/custom_section.wast#L105

This simplifies users, as it means that when they're only expecting one module, they don't have to handle the case where there's another module after it.

This PR also simplifies the read function to return a plain &ParserState rather than an Option<&ParserState> since ParserState already has EndWasm and Error to indicate that parsing should terminate. This seems like a useful simplification, though I'm definitely interested if anyone knows a better way to do this.


yurydelendik:
Thank you for the patch.

PR #13, opened 2017-06-06
Minor code tidying

sunfishcode:
This adds a few minor changes to use more idiomatic rust style.


yurydelendik:
Thank you for the patch.

Issue #22, opened 2017-06-22
Figure out how to add fuzz testing


yurydelendik:
More generic case of #21

We will need repeatable way to run fuzzing. Help is needed to add this to the project (e.g. add run-fuzz.sh utility).

/cc @frewsxcv

frewsxcv:
I can open a PR for this later today

sunfishcode:
With #23 merged, can this issue be closed?

yurydelendik:
> With #23 merged, can this issue be closed?

Yes, thanks.

PR #23, opened 2017-06-23
Add cargo-fuzz related files.

frewsxcv:
https://github.com/yurydelendik/wasmparser.rs/issues/22

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,3 @@
+target
+corpus
+artifacts
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "wasmparser-fuzz"
+version = "0.0.1"
+authors = ["Yury Delendik <ydelendik@mozilla.com>"]
+publish = false
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies.wasmparser]
+path = ".."
+
+[dependencies.libfuzzer-sys]
+git = "https://github.com/rust-fuzz/libfuzzer-sys.git"
+
+# Prevent this from interfering with workspaces
+[workspace]
+members = ["."]
+
+[[bin]]
+name = "parse"
+path = "fuzz_targets/parse.rs"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ keywords = ["parser", "WebAssembly", "wasm"]
 description = """
 A simple event-driven library for parsing WebAssembly binary files.
 """
+exclude = ["fuzz/"]
 
 [dependencies]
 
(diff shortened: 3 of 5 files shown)


frewsxcv:
To use this:

`cargo install cargo-fuzz`

`cargo fuzz run <fuzz target name>`, in this case `cargo fuzz run parse`

if you want to add some files to use as seeds for the fuzzer, add them to `fuzz/corpus/parse/`. i recommend adding all the files from `test/` into there and then rerunning

fuzz/.gitignore
```diff
@@ -0,0 +1,4 @@
+
```
yurydelendik: remove empty line

fuzz/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "wasmparser-fuzz"
+version = "0.0.1"
+authors = ["Automatically generated"]
```
yurydelendik: nit: fix author name

yurydelendik (changes requested):
Thank you.

I don't think we want to package fuzz into the crate package. Can you add exclude entry to the main Cargo.toml?

Can you squash commits? There is no reason to have "touch up" commits.

Please create a local README.md with the same instructions you provided in the PR comments. I was not able to run by following those instructions:

```
fuzz yury$ cargo fuzz run parse
error: could not decode the manifest file at "wasmparser.rs/wasmparser/Cargo.toml"
  caused by: redefinition of table `package` for key `package` at line 23
```

frewsxcv:
> Can you squash commits? There is no reason to have "touch up" commits.

https://github.com/blog/2141-squash-your-commits

> Please create a local README.md with the same instructions you provided in the PR comments. I was not able to run by following those instructions:

Which directory did you run the command from? You should run it from the repository root, not from within the `fuzz/` directory.

yurydelendik:
> https://github.com/blog/2141-squash-your-commits

Personally I'm trying not to do that: it removes signed commit verification and modifies origin of commit. I'm doing squashing only if contributors cannot to it himself. Would you like to squash it for you?

yurydelendik:
> You should run it from the repository root, not from within the fuzz/ directory.

Different error when run from repo's dir: https://gist.github.com/yurydelendik/c59534507de07801fed90865854874e9

frewsxcv:
Ah, you'll need to run nightly Rust for this. `rustup default nightly` if you're using rustup. I can add that to the README when I get around to it.

yurydelendik:
>  I can add that to the README when I get around to it.

Yes, please. Can you add fuzzing section to the README?

yurydelendik (approved):

frewsxcv:
Added a brief section to the README and squashed.

yurydelendik:
Thank you for the patch

PR #27, opened 2017-08-31
Add a `validate` function for convenient one-off byte sequence validation.

sunfishcode:
I wrote a simple `validate` helper function for a different project, which just wraps a ValidatingParser. It may be useful for others as well. It's analogous to [`WebAssembly.validate`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/validate) in the WebAssembly JS API.


yurydelendik:
Thank you for the patch.

PR #28, opened 2017-08-31
Add a public BinaryReader::read_local_decl() method.

stoklund:
This makes it possible to use the BinaryReader for also parsing the
local declarations that precede the function body.

I did not make `read_type()` public after all since it looks like it accepts some types that are not permitted for local variables, and it's possible we want to fix that in the future.


yurydelendik:
Thank you for the patch

PR #32, opened 2017-09-06
Changes in the API/states

### LINKED ISSUE #30: abort on custom linking section
On this wasm file:
```
$ hexdump -C fib.wasm 
00000000  00 61 73 6d 01 00 00 00  01 86 80 80 80 00 01 60  |.asm...........`|
00000010  01 7f 01 7f 03 82 80 80  80 00 01 00 04 84 80 80  |................|
00000020  80 00 01 70 00 00 05 83  80 80 80 00 01 00 00 07  |...p............|
00000030  8d 80 80 80 00 01 09 66  69 62 6f 6e 61 63 63 69  |.......fibonacci|
00000040  00 00 0a bc 80 80 80 00  01 3a 01 04 7f 41 01 21  |.........:...A.!|
00000050  01 02 40 20 00 41 01 48  0d 00 41 00 21 02 03 40  |..@ .A.H..A.!..@|
00000060  20 02 20 01 6a 21 03 20  01 21 04 20 01 21 02 20  | . .j!. .!. .!. |
00000070  03 21 01 20 00 41 7f 6a  22 00 0d 00 0b 20 04 0f  |.!. .A.j".... ..|
00000080  0b 41 00 0b 00 97 80 80  80 00 04 6e 61 6d 65 01  |.A.........name.|
00000090  8c 80 80 80 00 01 00 09  66 69 62 6f 6e 61 63 63  |........fibonacc|
000000a0  69 00 88 80 80 80 00 07  6c 69 6e 6b 69 6e 67     |i.......linking|
000000af
```

which contains a custom "linking" section, wasmparser's `dump` example program aborts with this output:
```
[…]
EndFunctionBody
EndSection
BeginSection { code: Custom { name: [110, 97, 109, 101], kind: Name }, range: Range { start: 143, end: 161 } }
NameSectionEntry(Function([Naming { index: 0, name: [102, 105, 98, 111, 110, 97, 99, 99, 105] }]))
EndSection
BeginSection { code: Custom { name: [108, 105, 110, 107, 105, 110, 103], kind: Linking }, range: Range { start: 175, end: 175 } }
thread 'main' panicked at 'Error: BinaryReaderError { message: "Unexpected EOF", offset: 175 }', examples/dump.rs:49:39
```

### LINKED ISSUE #31: validation error on large data sections
In modules with large data sections, such as the tanks or AngryBots demos, wasmparser issues a validation error:
```
wrong content in the data section: Error(BinaryReaderError { message: "string size in out of bounds", offset: 11423198 })
```

This appears to be because `read_data_entry_body` uses `read_string`, which requires that the size be less than MAX_WASM_STRING_SIZE. I believe this is incorrect; for example, SpiderMonkey's validator does not enforce this limit here.

yurydelendik:
Changes:

* Fixes #31 - data segment is read in chunks
* Fixes #30 - does not read custom sections by default
* Skips end markers after skipping function/section body

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmparser"
-version = "0.9.1"
+version = "0.9.2"
 authors = ["Yury Delendik <ydelendik@mozilla.com>"]
 license = "Apache-2.0"
 repository = "https://github.com/yurydelendik/wasmparser.rs"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmparser"
-version = "0.9.2"
+version = "0.9.3"
 authors = ["Yury Delendik <ydelendik@mozilla.com>"]
 license = "Apache-2.0"
 repository = "https://github.com/yurydelendik/wasmparser.rs"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmparser"
-version = "0.9.0"
+version = "0.9.1"
 authors = ["Yury Delendik <ydelendik@mozilla.com>"]
 license = "Apache-2.0"
 repository = "https://github.com/yurydelendik/wasmparser.rs"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmparser"
-version = "0.8.2"
+version = "0.9.0"
 authors = ["Yury Delendik <ydelendik@mozilla.com>"]
 license = "Apache-2.0"
 repository = "https://github.com/yurydelendik/wasmparser.rs"
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -22,6 +22,8 @@ use limits::{MAX_WASM_FUNCTION_LOCALS, MAX_WASM_FUNCTION_PARAMS, MAX_WASM_FUNCTI
 
 const MAX_WASM_BR_TABLE_SIZE: usize = MAX_WASM_FUNCTION_SIZE;
 
+const DATA_CHUNK_SIZE: u32 = MAX_WASM_STRING_SIZE as u32;
+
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wasmparser"
-version = "0.8.2"
+version = "0.9.2"
```
sunfishcode: It looks like the change to the behavior in `SkippingFunctionBody` would break existing users, so it should technically get a major-version bump.

src/parser.rs
```diff
@@ -1156,8 +1159,10 @@ pub enum ParserState<'a> {
     EndElementSectionEntry,
 
     BeginDataSectionEntry(u32),
-    DataSectionEntryBody(&'a [u8]),
     EndDataSectionEntry,
+    BeginDataSectionEntryBody(u32),
+    DataSectionEntryBodyChunk(&'a [u8]),
```
sunfishcode, line 1153: What's the reason for breaking up data section entries into chunks, instead of just returning one big slice? As a user of the library, I can always chunk up a big slice myself if I need to, and if I need the data to be contiguous for some reason, getting one big slice means I don't have to copy the data.

yurydelendik: I want to be able to change it to the "streaming" parser when end of the file is "unknown". Also feeding really big/invalid section size in input is a cause for OOM, fuzzers find those quickly.

sunfishcode: Ah, makes sense.

sunfishcode (approved):
With the SkippingFunctionBody change and the chunking changes, this version isn't backwards compatible, so I recommend a major-version bump.

Cargo.toml
yurydelendik: yes, I reserved 0.x as experimental version. After 1.x, I'll start semantic versioning.

sunfishcode:
I was mistaken; the [semver spec](http://semver.org/) does indeed say that 0.y.z is reserved for early development where anything can break. So no need to bump the major version!

PR #41, opened 2017-10-05
Adds threads operators support

yurydelendik:
TODO:
- [x] create tests


sunfishcode:
Looks good! I think we do want to add a preference to disable support for threads. I think it makes sense to put it in the validator so that we handle atomic opcodes the same way as invalid encodings.

src/parser.rs
```diff
@@ -466,6 +467,80 @@ pub enum Operator<'a> {
...
     I64TruncSSatF64,
     I64TruncUSatF64,
+
+    // 0xFE operators
+    // https://github.com/WebAssembly/threads/blob/master/proposals/threads/Overview.md
+    I32Extend8S,
+    I32Extend16S,
+    I64Extend8S,
+    I64Extend16S,
+    I64Extend32S,
```
sunfishcode: Can you adjust the comment here to make it clear that the Extend operators are not 0xFE operators?

yurydelendik: Moved them in the section above

sunfishcode:
Looks good to me!

PR #42, opened 2017-10-06
Add proposed Non-trapping Float-to-int Conversions

yurydelendik:
https://github.com/WebAssembly/nontrapping-float-to-int-conversions/blob/master/proposals/nontrapping-float-to-int-conversion/Overview.md#encoding


sunfishcode:
Looks good to me!

PR #43, opened 2017-10-11
Add lifetime parameters to &ParserState return values.

sunfishcode:
This change allows library users to keep ParserStates and the data in them alive longer than the Parser itself. For example, if a user wants to do deferred parsing of function bodies, they may want to keep a reference to the slice of the data representing the function body live after the main `Parser` is finished, so that they can construct a `BinaryReader` for it and decode it at some later time.


yurydelendik:
Thank you for the patch.

PR #44, opened 2018-01-16
Fix markdown syntax to be compatible with the pulldown parser.

sunfishcode:
This fixes warnings like the following when doing "cargo doc":

WARNING: documentation for this crate may be rendered differently using the new Pulldown renderer.
    See https://github.com/rust-lang/rust/issues/44229 for details.
WARNING: rendering difference in `External types as defined at https://webassembly.g ... .html#external-types`
   --> src/parser.rs:104:0
    /html[0]/body[1]/p[0] Text differs:
        expected: `... https://webassembly.github.io/spec/syntax/types.html#external-types`
        found:    `...`

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -41,7 +41,9 @@ pub enum CustomSectionKind {
     Linking,
 }
 
-/// Section code as defined at https://webassembly.github.io/spec/binary/modules.html#sections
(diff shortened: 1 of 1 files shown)


yurydelendik:
Thank you for the patch.

PR #46, opened 2018-03-09
Remove the size field from BrTable.

sunfishcode:
This shrinks the size of Operator from 32 bytes to 24 bytes. It
does mean that `read_table` can't use `with_capacity`, however
performance-sensitive uses can use BrTableIterator to process the
table elements to avoid allocating a Vec altogether.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -154,7 +154,6 @@ pub struct MemoryImmediate {
 /// A br_table entries representation.
 #[derive(Debug)]
 pub struct BrTable<'a> {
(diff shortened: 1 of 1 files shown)


yurydelendik (approved):
Looks good. (I need to bump version)

yurydelendik:
Merged at https://github.com/yurydelendik/wasmparser.rs/commit/5f2e19d7fff96eb00f1a6dd5e6b555599862b175

yurydelendik:
Thank you for the patch!

PR #47, opened 2018-03-12
Allow `BinaryReader` to be cloned, and shrink it slightly

sunfishcode:
I'm experimenting with a wasm translation optimization where I want to be able to look at the next opcode, without necessarily having my `BinaryReader` consume it. Allowing `BinaryReader` to be cloned seems like a simple way to make that possible, as I can just make a backup copy, read ahead, and then reset to my backup if I need to.

And while here, I noticed an opportunity to shrink `BinaryReader` slightly, though I can split this out into a separate PR if you'd like.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -577,6 +577,7 @@ enum InitExpressionContinuation {
 }
 
 /// A binary reader of the WebAssembly structures and types.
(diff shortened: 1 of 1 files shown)


yurydelendik:
thank you for the patch

PR #49, opened 2018-03-20
Add no_std support.

sunfishcode:
This adds `no_std` support to wasmparser.rs.

@eternaleye I've attempted to follow your advice in https://github.com/Cretonne/cretonne/pull/224 and use features additively, so I'm curious if you have any comments. wasmparser.rs uses `HashSet` internally, so to enable use of `hashmap_core` in an additive way, I've also created a `no_std` feature which enables that. You can enable both `std` and `no_std` at the same time, and it behaves like a `std` build. But you have to enable at least one of the two.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,5 +12,17 @@ exclude = ["fuzz/"]
 
 [dependencies]
 
+[dependencies.hashmap_core]
+version = "0.1.1"
+optional = true
+
 [badges]
 travis-ci = { repository = "yurydelendik/wasmparser.rs" }
+
+[features]
+# The "std" feature enables use of libstd. The "no_std" feature enables use
+# of some minimal std-like replacement libraries. At least one of these two
+# features needs to be enabled.
+default = ["std"]
+std = []
+no_std = ["hashmap_core"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,9 +20,9 @@ optional = true
 travis-ci = { repository = "yurydelendik/wasmparser.rs" }
 
 [features]
-# The "std" feature enables use of libstd. The "no_std" feature enables use
+# The "std" feature enables use of libstd. The "core" feature enables use
 # of some minimal std-like replacement libraries. At least one of these two
 # features needs to be enabled.
 default = ["std"]
 std = []
-no_std = ["hashmap_core"]
+core = ["hashmap_core"]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,6 +23,16 @@
 //! this is not the right library for you. You could however, build such
 //! a data-structure using this library.
 
+#![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr(not(feature = "std"), feature(alloc))]
+
+#[cfg(not(feature = "std"))]
+extern crate hashmap_core;
+
+#[cfg(not(feature = "std"))]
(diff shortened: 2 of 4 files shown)


eternaleye:
Ah, that's not _quite_ what I'd meant. :stuck_out_tongue: 

The "purely additive" property is about the _exposed API surface_; if `HashSet` is used only internally, then the `no_std` feature is unnecessary - simply use `hashmap_core` if the `std` feature is disabled.

To explain a bit, there are the following possibilities:

1. Enabling (or disabling) the feature leaves the API surface unchanged
    - This describes "purely internal" use, such as `HashSet` in this case
2. Enabling (resp. disabling) the feature expands (resp. shrinks) the API surface
    - For example, an `async` feature adding a function returning a `Future`
3. Enabling (or disabling) the feature replaces an element of the API surface
    - This would cover a case where a public function _took_ or _returned_ a `HashSet`, and whether the type came from `std` or `hashmap_core` was controlled by a feature
4. Enabling (resp. disabling) the feature shrinks (resp. expands) the API surface
    - This would cover a `no_std` feature that, say, did away with methods that return `Vec`

1 and 2 are kosher; 3 and 4 are not.

sunfishcode:
@eternaleye The problem that I ran into is that I couldn't find a way to tell Cargo to add hashmap_core as a dependency in non-`std` builds and *not* in `std` builds. Do you know of a way to do that?

eternaleye:
@sunfishcode:

```TOML
[target.'cfg(not(feature = "std"))'.dependencies]
hashmap_core = "0.1.2"
```

sunfishcode:
@eternaleye When I try that, it pulls in hashmap_core for `std` builds too.

```sh
$ git diff
diff --git a/Cargo.toml b/Cargo.toml
index 08bc406..aead01a 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,17 +12,12 @@ exclude = ["fuzz/"]
 
 [dependencies]
 
-[dependencies.hashmap_core]
-version = "0.1.1"
-optional = true
+[target.'cfg(not(feature = "std"))'.dependencies]
+hashmap_core = { version = "0.1.2" }
 
 [badges]
 travis-ci = { repository = "yurydelendik/wasmparser.rs" }
 
 [features]
-# The "std" feature enables use of libstd. The "no_std" feature enables use
-# of some minimal std-like replacement libraries. At least one of these two
-# features needs to be enabled.
 default = ["std"]
 std = []
-no_std = ["hashmap_core"]
$ cargo build
   Compiling hashmap_core v0.1.2
[...]
```
We specifically want to avoid pulling in hashmap_core in `std` builds.

eternaleye:
Ugh, looks like I was mistaken, and there's [no way to do that currently](https://github.com/rust-lang/cargo/issues/1839).

I _would_ suggest renaming the `no_std` feature to `core`, though. Semantically, it "adds support for a `core`-based environment" rather than negating the `std` option.

sunfishcode:
@eternaleye Sounds good, I've renamed `no_std` to `core` now. Thanks for your help!

sunfishcode:
Ok, I'm taking the liberty of merging this. Even though `no_std` mode is admittedly a little bit black magic right now, it doesn't affect users that don't enable it, and it's not *that* much clutter in the code. If it turns out to cause problems for anyone, I'd be willing to revert it.

PR #53, opened 2018-06-07
Rename memory instructions.

sunfishcode:
The wasm instructions `grow_memory` and `current_memory` were renamed to
`memory.grow` and `memory.size`, respectively:

https://github.com/WebAssembly/spec/pull/720

The patch here renames wasmparser.rs's `GrowMemory` and `CurrentMemory`
to `MemoryGrow` and `MemorySize`, to better align with the spec.


yurydelendik:
Thank you for the patch

PR #57, opened 2018-09-27
Format with the current stable rustfmt-preview.

sunfishcode:


yurydelendik:
check-rustfmt.sh needs to be updated?

yurydelendik:
can you also change Cargo.toml version?

sunfishcode:
In Cranelift, we ended up just folding check-rustfmt.sh into test-all.sh, so I've now synced these scripts to correspond to what we have there. Let me know if there's anything here that doesn't make sense.

And the Cargo.toml version is now updated.

sunfishcode:
I should mention that these changes also correspond to the upstream switch from the "old" rustfmt which was installed via cargo to the "new" rustfmt-preview which is installed via rustup. The names are a little confusing; "rustfmt-preview" is the name of the version that's included in the current stable release.

yurydelendik (approved):
looks good

PR #59, opened 2018-09-27
Implement anyref etc. from the reference-types proposal

sunfishcode:
This implements anyref, ref.null, and ref.isnull, from https://github.com/WebAssembly/reference-types.

This doesn't add any tests yet, because wabt doesn't yet support these types, but we'll add those tests when we can, and we'll test this with the SpiderMonkey testsuite once we hook everything up.


yurydelendik (approved):
Looks good with the library version change

PR #60, opened 2018-09-27
Fix various clippy warnings

sunfishcode:
This doesn't fix all the clippy warnings, just a rough pass through them. Some of these are subjective, so feel free to disagree; the fixes are split up into separate patches to make it easy to drop individual ones.


yurydelendik:
Thank you for the patch.

PR #61, opened 2018-09-27
Miscellaneous repo bits

sunfishcode:
This adds in a few things from the cranelift repo that we've found useful there.


yurydelendik:
Thank you

PR #63, opened 2018-09-27
Use Rust's str::from_utf for UTF-8 validation.

sunfishcode:
This replaces the reference-implementation UTF-8 decoder with Rust's optimized standard library version.


yurydelendik:
Thank you

PR #65, opened 2018-10-10
Move BinaryReader out of parser.rs

yurydelendik:


src/parser.rs
```diff
@@ -15,72 +15,21 @@
...
-    pub message: &'static str,
-    pub offset: usize,
-}
-
-pub type Result<T> = result::Result<T, BinaryReaderError>;
+pub use primitives::{
+    BinaryReaderError, BrTable, CustomSectionKind, ExternalKind, FuncType, GlobalType, Ieee32,
+    Ieee64, MemoryImmediate, MemoryType, Naming, Operator, ResizableLimits, Result, SectionCode,
+    TableType, Type,
+};
```
sunfishcode, line 27: Rather than having parser.rs do `pub use` like this, I think it would be simpler to just have lib.rs do `pub use primitives::BinaryReaderError;` and so on, using them directly from primitives.rs rather than going through parser.rs.

That way, parser.rs can also do `use {BinaryReaderError, CustomSectionKind, ...};`, and only use the things that it itself needs.

sunfishcode (commented):
Nice refactoring!

sunfishcode (approved):

PR #66, opened 2018-10-12
[WIP]  Attempt to collect all states field into single structure 

yurydelendik:
Always wanted to do that to make Parser serializable.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -195,16 +195,39 @@ pub trait WasmDecoder<'a> {
     fn last_state(&self) -> &ParserState<'a>;
 }
 
+#[derive(Debug, Copy, Clone)]
+struct WasmSectionBounds<'a> {
+    code: SectionCode<'a>,
+    range: Range,
+}
+
+#[derive(Debug, Copy, Clone)]
+struct WasmSectionEntryBounds<'a> {
+    section_bounds: WasmSectionBounds<'a>,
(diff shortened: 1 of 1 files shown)


sunfishcode:
When I look at `get_section_entries_left`, it feels awkward that we have to do this `match` every time we read an entry -- actually twice per entry, once to read it and once to update it.

As a thought experiment, what would you think about an API that worked something like this:

 - `Parser` returns a state saying we're in the "type" section, for example.
 - The user then chooses either `parser.skip_section()` or `parser.get_type_reader()`.
 - `parser.get_type_reader()` would return a `TypeReader` which is similar to `BinaryReader` but which just has a `read_type_entry()` function. The `section_entries_left` field would be tracked inside the `TypeReader`.

Similar things could apply to other sections as well.

This would move away from the idea of a wasm module as just a sequence of tokens that one can step all the way through with a single loop. From Cranelift's perspective, this would be fine, because we're already not using a single loop :-). But I'm interested in what you think of this.

yurydelendik:
> This would move away from the idea of a wasm module as just a sequence of tokens that one can step all the way through with a single loop.  ... But I'm interested in what you think of this.

Yes, I'm looking at this as well (I will open different PR soon). I just want to verify effectiveness of rust compiler, and monitor generated code performance for current architecture.

yurydelendik:
Closing in favor of #68

PR #67, opened 2018-10-15
Implement ModuleReader and sections readers

yurydelendik:


src/readers/type_section.rs
```diff
@@ -0,0 +1,107 @@
...
+    reader: BinaryReader<'a>,
+    state: TypeSectionReaderState,
+}
+
+impl<'a> TypeSectionReader<'a> {
+    pub fn new(data: &'a [u8]) -> TypeSectionReader<'a> {
+        TypeSectionReader {
+            reader: BinaryReader::new(data),
+            state: TypeSectionReaderState::Initial,
+        }
```
sunfishcode: I like where this is going, and my suggestion is to take it a step further:
 - give `TypeSectionReader` a `u32` count member, and have it read the count right here in `TypeSectionReader::new`, which would then return `Result<TypeSectionReader<'a>>`
 - make `read_entry` return a `Result<FuncType>` rather than assigning to `self.state`
 - eliminate the `state` member

The usage could then look like this:
```rust
   let type_reader = reader.get_type_section_reader()?;
   let count = type_reader.num_entries();
   for index in 0..count {
        let ty = type_reader.read_entry()?;
        ... // use ty, index, and count as needed
   }
```

Does that look reasonable?

src/readers/module.rs
```diff
@@ -0,0 +1,199 @@
...
+    // the type section.
+    pub fn get_type_section_reader(&self) -> TypeSectionReader {
+        match self.state {
+            ModuleReaderState::Section {
+                code: SectionCode::Type,
+                range,
+            } => TypeSectionReader::new(range.slice(self.reader.buffer)),
+            _ => panic!("Invalid state for get_type_section_reader"),
+        }
+    }
```
sunfishcode: Should `get_type_section_reader` also advance self.reader so that it's ready to read the following section once the user has finished with the type section?

src/readers/type_section.rs
sunfishcode: In fact, another step yet would be to do a similar thing at the `ModuleReader` level. If we made `ModuleReader::new` read the wasm version, and `get_type_section_reader` return an `Option` with `None` indicating that there is no type section, the user API could look like this:

```rust
    fn read_wasm(data: &[u8]) -> Result<()> {
       let (reader, version) = ModuleReader::new(data)?;
       // use version
       reader.skip_custom_sections();
       if let Some(type_reader) = reader.get_type_section_reader()? {
           // use type_reader
       }
       reader.skip_custom_sections();
       if let Some(import_reader) = reader.get_import_section_reader()? {
           // use import_reader
       }
       // and so on
    }
```
I did `skip_custom_sections()` here to keep this example simple, but there could be a way to handle custom sections at those points too.

It is a tradeoff; this overall approach does require the user to know a bit more about wasm's structure. However, using wasmparser in cranelift, I've found that it has to know some of this anyway, because we can't process some sections until we've seen other ones.

src/readers/type_section.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// # let data: &[u8] = &[0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00,
+    /// #     0x01, 0x4, 0x01, 0x60, 0x00, 0x00, 0x03, 0x02, 0x01, 0x00,
+    /// #     0x0a, 0x05, 0x01, 0x03, 0x00, 0x01, 0x0b];
+    /// use wasmparser::ModuleReader;
+    /// let mut reader = ModuleReader::new(data).expect("module reader");
+    /// let section = reader.read().expect("section");
+    /// let mut type_reader = section.get_type_section_reader().expect("type section reader");
+    /// assert!(type_reader.get_count() == 1);
+    /// let ty = type_reader.read().expect("type #1");
+    /// println!("Type {:?}", ty);
```
sunfishcode: I suggest using a loop in the code example here to show the expected way to use this API:
```rust
          for index in 0..type_reader.get_count() {
              let ty = type.reader().read().expect("read a type");
              println!("Type {}: {:?}", index, ty);
          }
```

src/readers/type_section.rs
```diff
@@ -0,0 +1,52 @@
...
+        let mut reader = BinaryReader::new(data);
+        let count = reader.read_var_u32()?;
+        Ok(TypeSectionReader { reader, count })
+    }
+
+    pub fn get_count(&self) -> u32 {
+        self.count
+    }
+
+    /// Reads content of the type section. See also `TypeSectionReaderState` enum.
```
sunfishcode: Stale comment about `TypeSectionReaderState`.

src/readers/module.rs
```diff
@@ -0,0 +1,114 @@
...
+        self.version
+    }
+
+    pub fn eof(&self) -> bool {
+        self.read_ahead.is_none() && self.reader.eof()
+    }
+
+    /// Reads next top-level record from the WebAssembly binary data.
+    /// The methods returns reference to current state of the reader.
+    /// See `ModuleReaderState` enum.
```
sunfishcode: Stale comment about `ModuleReaderState`.

src/readers/module.rs
```diff
@@ -0,0 +1,114 @@
...
+            return Err(BinaryReaderError {
+                message: "Section header is too big to fit into section body",
+                offset: payload_end,
+            });
+        }
+        let body_start = self.reader.position;
+        self.reader.skip_to(payload_end);
+        Ok(Section {
+            code,
+            offset: self.reader.position,
```
sunfishcode: Is passing the `self.reader.position` after doing the `skip_to` intentional? It's not clear to me what the purpose of this offset field is.

src/readers/module.rs
```diff
@@ -0,0 +1,114 @@
...
+impl<'a> Section<'a> {
+    // Creates reader for the type section. Available when the reader just read
+    // the type section.
+    pub fn get_type_section_reader(&self) -> Result<TypeSectionReader> {
+        match self.code {
+            SectionCode::Type => TypeSectionReader::new(self.data),
+            _ => panic!("Invalid state for get_type_section_reader"),
+        }
+    }
+}
```
sunfishcode, line 165: We'll need an accessor to read the `code` field, so that we can tell what kind of section it is.

sunfishcode (commented):
Cool, this looks good!

PR #75, opened 2018-11-06
Add a way to build a `FunctionBody` using a raw byte buffer.

sunfishcode:
This will be useful in SpiderMonkey where we decode most of the module up front, and then hand off parts of the code section to worker threads for compilation, so they just want to construct a `FunctionBody` and then start decoding.


yurydelendik:
Thank you for the patch.

PR #76, opened 2018-11-08
Add operator reading with offsets

yurydelendik:


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wasmparser"
-version = "0.21.8"
+version = "0.21.9"
```
sunfishcode, line 3: This should be a minor version bump, as it adds new features.

src/readers/operators.rs
```diff
@@ -49,6 +49,23 @@ impl<'a> OperatorsReader<'a> {
...
+    where
+        'a: 'b,
+    {
+        OperatorsIteratorWithOffsets {
+            reader: self,
+            err: false,
+        }
+    }
+
+    pub fn read_with_offset<'b>(&mut self) -> Result<(usize, Operator<'b>)>
```
sunfishcode: This is just aesthetic preference, but I'd find this a little more natural if the `Operator<'b>` came first in the tuple, with the `usize` second.

src/readers/operators.rs
```diff
@@ -104,3 +121,45 @@ impl<'a> Iterator for OperatorsIterator<'a> {
...
+    ///         ops
+    ///     );
+    /// }
+    /// ```
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.err || self.reader.eof() {
+            return None;
+        }
+        let result = self.reader.read_with_offset();
+        self.err = result.is_err();
```
sunfishcode, line 163: Random idea: If `BinaryReader` had a way to say "skip to the end", then we could change this code to
```rust
    if result.is_err() {
        self.reader.skip_to_end();
    }
```
and then we wouldn't need these `self.err` fields in the various reader classes that have them, and less branching when calling `next`. Does that sound reasonable?

sunfishcode (approved):
Looks good!

Cargo.toml
yurydelendik: AFAIK that's a minor bump for "0.x.y" versions

src/readers/operators.rs
yurydelendik: I tested this approach -- it resulted in 5% speed loss. I moved the commit into the [skip_to_end](https://github.com/yurydelendik/wasmparser.rs/tree/skip_to_end) branch.

PR #78, opened 2018-11-18
handle integer overflows in polymorphic stack validation

camshaft:
I ran into another crash with fuzzing. I included the file in `tests/invalid`. Let me know if there's a better place to put it (it looks like all of those tests are generated from the official testsuite)


yurydelendik (approved):
Thank you for the patch!

PR #81, opened 2018-12-12
Export OperatorValidatorConfig so that wasmparser users can use it.

sunfishcode:
`validate` is exported, and takes an `OperatorValidatorConfig` to configure it, so make `OperatorValidatorConfig` available too.


yurydelendik:
Thank you for the patch

PR #82, opened 2018-12-12
Implement mutable global imports.

sunfishcode:
This adds support for mutable globals in the validator, as described here: https://github.com/WebAssembly/mutable-global/

Also, export `ValidatingParserConfig` too, since that's also needed.


yurydelendik:
Thank you for the patch

PR #83, opened 2019-01-24
Minor updates found when integrating into wasm-bindgen tooling

alexcrichton:
This is a few minor updates that I ran into when preparing https://github.com/rustwasm/walrus/pull/26, a crate for eventually powering wasm-bindgen tooling. Most of it's pretty straightforward, but I'm particularly curious what you think about the change of signature in `read_string`!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmparser"
-version = "0.23.0"
+version = "0.24.0"
 authors = ["Yury Delendik <ydelendik@mozilla.com>"]
 license = "Apache-2.0 WITH LLVM-exception"
 repository = "https://github.com/yurydelendik/wasmparser.rs"
--- a/examples/dump.rs
+++ b/examples/dump.rs
@@ -9,10 +9,6 @@ use wasmparser::Parser;
 use wasmparser::ParserState;
 use wasmparser::WasmDecoder;
 
-fn get_name(bytes: &[u8]) -> &str {
-    str::from_utf8(bytes).ok().unwrap()
-}
-
 fn main() {
     let args = env::args().collect::<Vec<_>>();
     if args.len() != 2 {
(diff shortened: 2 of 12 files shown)


yurydelendik:
> I'm particularly curious what you think about the change of signature in `read_string`

I chose `&[u8]` instead of `&str` only because,  the WebAssembly is/was inclined to use random set of bytes as names, e.g. see http://webassembly.github.io/spec/core/binary/modules.html#custom-section . That said, it looks like this patch is correct for http://webassembly.github.io/spec/core/binary/values.html#binary-name fields, e.g. in the import entries, but cannot be used for e.g. custom section name. I'll try to review and match utf-8 names reading to the spec recommendations.

alexcrichton:
Ok sounds good to me! I was thinking yeah for any non-spec-stated areas you'd get a byte slice, but wherever the spec mandates utf-8 we'd return `&str`

sunfishcode:
That looks like an error in the spec. The spec testsuite specifically tests that non-UTF-8 custom section names are rejected: https://github.com/WebAssembly/spec/blob/master/test/core/utf8-custom-section-id.wast.

sunfishcode:
On further reading, the spec is correct; it says the section name is a [name](http://webassembly.github.io/spec/core/syntax/values.html#names), which is a UTF-8 encoded string. The part about "uninterpreted bytes applies to the body of the custom section. So we should be able to use `str` for the custom section name.

yurydelendik:
>  it says the section name is a name,

my bad, yep, that patch looks correct in this case. @sunfishcode any objections on changing API from cranelift point of view?

sunfishcode:
`str` sounds good. We can always call `as_bytes()` on the `str` if we ever need it to be a `&[u8]` :-).

yurydelendik:
Thank you for the patch!

PR #87, opened 2019-01-29
Add experimental support for bulk memory operations

alexcrichton:
This commit adds support for the experimental bulk memory operations
proposal at https://github.com/WebAssembly/bulk-memory-operations. This
implements the proposal largely to the best extent we can, but in
particular passive element segments remain somewhat questionable because
their initializers are supposed to be opcodes but the encodings there
haven't been settled with the upstream reference types proposal just
yet.


src/readers/data_section.rs
```diff
@@ -75,30 +83,37 @@ impl<'a> DataSectionReader<'a> {
...
-        let memory_index = self.reader.read_var_u32()?;
-        let init_expr = {
-            let expr_offset = self.reader.position;
-            self.reader.skip_init_expr()?;
-            let data = &self.reader.buffer[expr_offset..self.reader.position];
-            InitExpr::new(data, self.reader.original_offset + expr_offset)
+        let flags = self.reader.read_var_u32()?;
+        let kind = if flags == 1 {
+            DataKind::Passive
+        } else {
```
yurydelendik, line 100: the spec defines only 0, 1, and 2 as acceptable values: do we want to return Err here if it is not.

src/readers/element_section.rs
```diff
@@ -154,12 +165,26 @@ impl<'a> ElementSectionReader<'a> {
...
-        let init_expr = {
-            let expr_offset = self.reader.position;
-            self.reader.skip_init_expr()?;
-            let data = &self.reader.buffer[expr_offset..self.reader.position];
-            InitExpr::new(data, self.reader.original_offset + expr_offset)
+        let flags = self.reader.read_var_u32()?;
+        let kind = if flags == 1 {
+            let ty = self.reader.read_type()?;
+            ElementKind::Passive(ty)
+        } else {
```
yurydelendik, line 172: the spec defines only 0, 1, and 2 as acceptable values: do we want to return Err here if it is not.

yurydelendik (approved):
Looks good with formatting fix and 0,1,2-validation.

alexcrichton:
Sure thing, updated!

yurydelendik:
Thank you for the patch!

PR #92, opened 2019-02-21
fix bug when branching to a loop

cdetrio:
Here is a test case that triggers the bug:

```
(module
  (type $t0 (func (result i32)))
  (func $main (export "main") (type $t0) (result i32)
    (block $B1 (result i32)
      (loop $L1 (result i32)
        (i32.const 77) ;; leave this val on the stack
        (i32.const 0) ;; conditional
        (br_if $L1) ;; branch to the loop with one value on the stack
        (br $B1)))))
```
I'm not sure if this is the best fix, but it passes all the tests.


yurydelendik:
Thank you for the patch!

PR #95, opened 2019-03-12
Add benchmark

### LINKED ISSUE #8: Figure out how to add [bench] benchmarks.
I tried doing this but it the test library is not stable.

iamonuwa:
<del>Fixes</del><ins>work towards</ins> #8

### DIFF
--- a/benches/benchmark.rs
+++ b/benches/benchmark.rs
@@ -1,5 +1,4 @@
 pub fn read_file_data(path: &PathBuf) -> Vec<u8> {
-    println!("Parsing {:?}", path);
     let mut data = Vec::new();
     let mut f = File::open(path).ok().unwrap();
     f.read_to_end(&mut data).unwrap();
--- a/benches/benchmark.rs
+++ b/benches/benchmark.rs
@@ -27,31 +27,35 @@ use std::io::Read;
 use std::path::PathBuf;
 
 fn it_works_benchmark(c: &mut Criterion) {
-    let mut data: Vec<u8> = vec![];
+    let mut data: Vec<Vec<u8>> = vec![vec![]];
     for entry in read_dir("tests").unwrap() {
         let dir = entry.unwrap();
         if !dir.file_type().unwrap().is_file() {
             continue;
         }
-        data = read_file_data(&dir.path());
+        data.push(read_file_data(&dir.path()));
     }
-    c.bench_function("it works benchmark", move |b| {
-        b.iter(|| Parser::new(data.as_slice()))
-    });
+    for d in data {
+        c.bench_function("it works benchmark", move |b| {
+            b.iter(|| Parser::new(d.as_slice()))
+        });
+    }
 }
 
 fn validator_not_fails_benchmark(c: &mut Criterion) {
-    let mut data: Vec<u8> = vec![];
+    let mut data: Vec<Vec<u8>> = vec![vec![]];
     for entry in read_dir("tests").unwrap() {
         let dir = entry.unwrap();
         if !dir.file_type().unwrap().is_file() {
             continue;
         }
-        data = read_file_data(&dir.path());
+        data.push(read_file_data(&dir.path()));
     }
-    c.bench_function("validator no fails benchmark", move |b| {
-        b.iter(|| ValidatingParser::new(data.as_slice(), VALIDATOR_CONFIG));
-    });
+    for d in data {
+        c.bench_function("validator no fails benchmark", move |b| {
+            b.iter(|| ValidatingParser::new(d.as_slice(), VALIDATOR_CONFIG));
+        });
+    }
 }
 
 criterion_group!(benchmark, it_works_benchmark, validator_not_fails_benchmark);
--- a/benches/benchmark.rs
+++ b/benches/benchmark.rs
@@ -27,29 +27,29 @@ use std::io::Read;
 use std::path::PathBuf;
 
 fn it_works_benchmark(c: &mut Criterion) {
-    c.bench_function("it works benchmark", move |b| {
-        for entry in read_dir("tests").unwrap() {
-            let dir = entry.unwrap();
-            if !dir.file_type().unwrap().is_file() {
-                continue;
-            }
-            let data = read_file_data(&dir.path());
-            b.iter(|| Parser::new(data.as_slice()))
+    for entry in read_dir("tests").unwrap() {
+        let dir = entry.unwrap();
+        if !dir.file_type().unwrap().is_file() {
+            continue;
         }
-    });
+        let data = read_file_data(&dir.path());
+        c.bench_function("it works benchmark", move |b| {
+            b.iter(|| Parser::new(data.as_slice()))
+        });
+    }
 }
 
 fn validator_not_fails_benchmark(c: &mut Criterion) {
-    c.bench_function("validator no fails benchmark", move |b| {
-        for entry in read_dir("tests").unwrap() {
-            let dir = entry.unwrap();
-            if !dir.file_type().unwrap().is_file() {
-                continue;
-            }
-            let data = read_file_data(&dir.path());
-            b.iter(|| ValidatingParser::new(data.as_slice(), VALIDATOR_CONFIG));
+    for entry in read_dir("tests").unwrap() {
+        let dir = entry.unwrap();
+        if !dir.file_type().unwrap().is_file() {
+            continue;
         }
-    });
+        let data = read_file_data(&dir.path());
+        c.bench_function("validator no fails benchmark", move |b| {
+            b.iter(|| ValidatingParser::new(data.as_slice(), VALIDATOR_CONFIG));
+        });
+    }
 }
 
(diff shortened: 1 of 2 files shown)


iamonuwa:
@mbebenita @yurydelendik please review

benches/benchmark.rs
```diff
@@ -0,0 +1,57 @@
+pub fn read_file_data(path: &PathBuf) -> Vec<u8> {
+    println!("Parsing {:?}", path);
```
yurydelendik: can you remove this line

yurydelendik (commented):
When I add sleep call to the read_file_data(), I see changes in the benchmark performance run -- any I/O shall not affect test results.

benches/benchmark.rs
iamonuwa: Okay

iamonuwa:
> When I add sleep call to the read_file_data(), I see changes in the benchmark performance run -- any I/O shall not affect test results.

Can you elaborate on this please?

yurydelendik:
> Can you elaborate on this please?

Benchmark test need to measure performance of the Parser/ValidatingParser. Looks like the solution is measuring the performance of the read_file_data function as well, which is not part of the Parser, and definitely skews the results. As I mentioned in [the comment](https://github.com/yurydelendik/wasmparser.rs/issues/8#issuecomment-471999627), to avoid that you may need to load/cache all wasm data prior running the testing.

iamonuwa:
> > Can you elaborate on this please?
> 
> Benchmark test need to measure performance of the Parser/ValidatingParser. Looks like the solution is measuring the performance of the read_file_data function as well, which is not part of the Parser, and definitely skews the results. As I mentioned in [the comment](https://github.com/yurydelendik/wasmparser.rs/issues/8#issuecomment-471999627), to avoid that you may need to load/cache all wasm data prior running the testing.

Any preferred option for in-memory cache?

yurydelendik:
> Any preferred option for in-memory cache?

`std::vec::Vec<>` shall work fine.

iamonuwa:
Alright, I'll work with this now

iamonuwa:
@yurydelendik If I get you correctly, the benchmark should run for each of the tests? Right

yurydelendik:
>  If I get you correctly, the benchmark should run for each of the tests? Right

I don't understand the question. But what I was saying that performance of read_file_data (which has I/O operations) shall not influence results of the benchmark test results. #8 is not clear on that, but bench shall test performances of the Parser (and ValidatingParser), and we want to have most of its functionality covered during test run. That said, the enough tests shall be executed during bench to get performance coverage of most of the code base. (The performances of reading files from disk is irrelevant)

iamonuwa:
Do you have a link or guide that can explain how to remove the additional functions? This is how I normally write benchmarks

iamonuwa:
This is how the output looks like.
image

I had to move the data outside the bench_function

iamonuwa:
Can you confirm?

yurydelendik:
Tested again: still a problem

```
Yurys-Workspace:wasmparser yury$ cargo bench
    Finished release [optimized] target(s) in 0.18s
     Running target/release/deps/wasmparser-b27d87a6221dce03

running 7 tests
test tests::simple_tests::default_read ... ignored
test tests::simple_tests::default_read_with_input ... ignored
test tests::simple_tests::it_works ... ignored
test tests::simple_tests::skipping ... ignored
test tests::simple_tests::validator_fails ... ignored
test tests::simple_tests::validator_not_fails ... ignored
test validator::test_validate ... ignored

test result: ok. 0 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out

     Running target/release/deps/benchmark-97097cd7f0076b61
Gnuplot not found, disabling plotting
it works benchmark      time:   [115.85 ns 118.91 ns 123.55 ns]                               
Found 8 outliers among 100 measurements (8.00%)
  8 (8.00%) high severe

validator no fails benchmark                                                                            
                        time:   [251.47 ns 256.82 ns 263.57 ns]
Found 9 outliers among 100 measurements (9.00%)
  2 (2.00%) high mild
  7 (7.00%) high severe

Gnuplot not found, disabling plotting
Yurys-Workspace:wasmparser yury$ vi benches/benchmark.rs 
Yurys-Workspace:wasmparser yury$ git diff
diff --git a/benches/benchmark.rs b/benches/benchmark.rs
index b1d600c..39282d0 100644
--- a/benches/benchmark.rs
+++ b/benches/benchmark.rs
@@ -1,6 +1,7 @@
 pub fn read_file_data(path: &PathBuf) -> Vec<u8> {
     let mut data = Vec::new();
     let mut f = File::open(path).ok().unwrap();
+::std::thread::sleep(::std::time::Duration::from_micros(50));
     f.read_to_end(&mut data).unwrap();
     data
 }
Yurys-Workspace:wasmparser yury$ cargo bench
   Compiling wasmparser v0.29.2 (/Users/yury/Work/junk/wasmparser.rs/wasmparser)
    Finished release [optimized] target(s) in 1.02s
     Running target/release/deps/wasmparser-b27d87a6221dce03

running 7 tests
test tests::simple_tests::default_read ... ignored
test tests::simple_tests::default_read_with_input ... ignored
test tests::simple_tests::it_works ... ignored
test tests::simple_tests::skipping ... ignored
test tests::simple_tests::validator_fails ... ignored
test tests::simple_tests::validator_not_fails ... ignored
test validator::test_validate ... ignored

test result: ok. 0 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out

     Running target/release/deps/benchmark-97097cd7f0076b61
Gnuplot not found, disabling plotting
it works benchmark      time:   [232.70 ns 260.96 ns 290.36 ns]                               
                        change: [+124.96% +143.21% +160.95%] (p = 0.00 < 0.05)
                        Performance has regressed.

validator no fails benchmark                                                                            
                        time:   [308.30 ns 347.51 ns 393.16 ns]
                        change: [+22.541% +31.809% +43.493%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 12 outliers among 100 measurements (12.00%)
  2 (2.00%) high mild
  10 (10.00%) high severe

Gnuplot not found, disabling plotting
```

Notice the performance "regressed" because of in slow down in reading of the file from disk.

yurydelendik:
> I had to move the data outside the bench_function

Did you commit this code?

iamonuwa:
Apologies, pushing new update now

benches/benchmark.rs
```diff
@@ -0,0 +1,56 @@
...
+use std::path::PathBuf;
+
+fn it_works_benchmark(c: &mut Criterion) {
+    for entry in read_dir("tests").unwrap() {
+        let dir = entry.unwrap();
+        if !dir.file_type().unwrap().is_file() {
+            continue;
+        }
+        let data = read_file_data(&dir.path());
+        c.bench_function("it works benchmark", move |b| {
```
yurydelendik: (here and below) move bench_function call out of this loop, store loaded file in the vector, add second loop in the closure and iterate on loaded files.

yurydelendik (changes requested):

benches/benchmark.rs
```diff
@@ -0,0 +1,58 @@
...
+
+use criterion::Criterion;
+use wasmparser::{OperatorValidatorConfig, Parser, ValidatingParser, ValidatingParserConfig};
+
+use std::fs::{read_dir, File};
+use std::io::Read;
+use std::path::PathBuf;
+
+fn it_works_benchmark(c: &mut Criterion) {
+    let mut data: Vec<u8> = vec![];
```
yurydelendik: Use `Vec<Vec<u8>>` here

iamonuwa: Using this causes failure on the bench

image

yurydelendik: See https://github.com/yurydelendik/wasmparser.rs/pull/95#discussion_r265293414

> ... store loaded file in the vector, add second loop in the closure and iterate on loaded files.

iamonuwa:
@yurydelendik @mbebenita  please review

benches/benchmark.rs
```diff
@@ -0,0 +1,62 @@
...
+fn it_works_benchmark(c: &mut Criterion) {
+    let mut data: Vec<Vec<u8>> = vec![vec![]];
+    for entry in read_dir("tests").unwrap() {
+        let dir = entry.unwrap();
+        if !dir.file_type().unwrap().is_file() {
+            continue;
+        }
+        data.push(read_file_data(&dir.path()));
+    }
+    for d in data {
```
yurydelendik: As pointed in https://github.com/yurydelendik/wasmparser.rs/pull/95/files#r265293414
> add second loop in the closure and iterate on loaded files.

Could you move this loop *inside* the bench_function ?

iamonuwa: Yeah sure!

iamonuwa:
@yurydelendik please review

yurydelendik (approved):

yurydelendik:
Thank you for the patch

iamonuwa:
Thank you for supporting!

PR #100, opened 2019-04-18
Validate function body

yurydelendik:
Making it possible to validate function code separate from entire module.

/cc @pchickey


yurydelendik:
The validate() performance improvement during tanks.wasm run: callgrind indicates Ir goes from 1,599,989,220 to 1,126,156,181, and speed of 260ms -> 164ms.

PR #101, opened 2019-05-29
Fix url

11Takanori:
Fix 404 Not Found Errors


yurydelendik:
Thank you for the patch.

PR #102, opened 2019-06-05
Use an explicit `dyn` for trait objects

fitzgen:
Silences a bunch of warnings.


yurydelendik:
Thank you for the patch.

PR #103, opened 2019-06-05
Multi value

fitzgen:
Builds on top of #102


fitzgen:
(Note: this just enables function calls with multiple return values, it doesn't attempt to handle multi-value blocks yet, which I figure can come in follow up PRs).

fitzgen:
Added support for multi-value blocks, which was slightly more involved since it had to mess with control frames / `BlockState` in the validator a bit.

yurydelendik:
It looks good. The only question have is where I can find a spec for the encoding used in the patch? There is https://webassembly.github.io/multi-value/core/binary/types.html#function-types, but it does not match the logic at the `read_type_or_func_type`.

fitzgen:
See https://webassembly.github.io/multi-value/core/binary/instructions.html#binary-blocktype and the paragraph just above it:

> Block types are encoded in special compressed form, by either the byte 0x40 indicating the empty type, as a single value type, or as a type index encoded as a positive signed integer.

src/binary_reader.rs
```diff
@@ -790,19 +790,30 @@ impl<'a> BinaryReader<'a> {
         })
     }
 
+    fn read_type_or_func_type(&mut self) -> Result<TypeOrFuncType> {
```
yurydelendik: let's change the method name to `read_block_type` to match the spec's description (`TypeOrFuncType` looks fine, I think)

src/binary_reader.rs
```diff
@@ -790,19 +790,30 @@ impl<'a> BinaryReader<'a> {
...
+        let position = self.position;
+        if let Ok(ty) = self.read_type() {
+            Ok(TypeOrFuncType::Type(ty))
+        } else {
+            self.position = position;
+            let idx = self.read_var_u32()?;
```
yurydelendik: We need to add a comment here that we are trying to read var `s33` number. (or maybe add read_var_s33_positive, which will replace error to "Invalid positive var_s33"?)

yurydelendik (commented):
Looks good with the minor changes

fitzgen:
Pushed a new commit with feedback addressed.

yurydelendik:
Thank you for the patch.

fitzgen:
Thanks!

PR #105, opened 2019-06-26
Fixed compilation when compiling with no std

syrusakbary:
When compiling with `features = ["core"]` (without std), there was an error.

This PR fixes it :)


yurydelendik:
Thank you for the patch.

PR #106, opened 2019-06-27
Update to Rust 2018 Edition

sunfishcode:
The main goal here is to fix a few `no_std` issues:
 - Fix some `no_std` compilation errors
 - Update no_std to use hashbrown rather than hashmap_core
 - Update no_std support to work more like Cranelift's no_std support works, which uses Rust 2018.


yurydelendik:
Thank you for the patch!

PR #111, opened 2019-06-29
fix error in benchmarks, added the required  field

Sreyas-Sreelal:
The `enable_multi_value` field was missing


yurydelendik:
Thank you for the patch.

PR #119, opened 2019-07-30
Fixes #118: Add support for atomic.fence

### LINKED ISSUE #118: Add support for Atomic fence
The opcode is 0xfe 0x03. It can be used even if there's no memory in the current module, since it can have an effect when e.g. it's inserted between calls to functions of another module that use shared memory.

bnjbvr:


yurydelendik:
thank you for the patch

PR #121, opened 2019-08-08
enable/disable floats configuration

YaronWittenstein:
Hi @yurydelendik 

I've added a new operator config option called `enable_floats` (defaults to `true`) under the `OperatorValidatorConfig`.

I believe this PR will be very useful for most (if not all) Blockchain projects using `wasm` for running smart-contracts.

Blockchain project using wasm avoid floats in order to stay deterministic across different hardware / operating-systems.

Thank you!
   Yaron


sunfishcode:
This patch appears to disable float operators, but not float types, for example in function signatures, locals, or globals. Is that intended?

If you do intend to disable float types, disabling float operators would then be unnecessary and the patch could be simpler, because it's not possible to use float operators if there are no float types (except in unreachable code, but there they can be ignored).

Also, while I don't object to the patch, I do want to mention for what it's worth that WebAssembly's floating-point is deterministic in all respects except for NaN bits. Some platforms tolerate some floating-point rounding nondeterminism, but [WebAssembly does not](https://github.com/WebAssembly/testsuite/blob/master/float_misc.wast#L122).

YaronWittenstein:
@sunfishcode Hi!

Basically, I wanted to make sure no floating operators will appear in the code.
I guess that floating types could still appear. But I'd rather avoid them as well.

I wasn't sure what's the best way to disable the float types in the source code...
(I didn't want to patch the `primitives.rs` file).

Thank you for the clarification regarding the deterministic nature of floats in wasm.
Nonetheless, I know that the company I work for would like to avoid floats feature-wise (and as said, I believe more Blockchain projects will find that feature relevant).

yurydelendik:
> I believe more Blockchain projects will find that feature relevant

Except float, which features will be beneficial? The reason why I'm asking is that it might be beneficial to move it to the feature level and use `#[cfg(feature = ` for blockchain-related subset. Perhaps even for deterministic stuff too?

YaronWittenstein:
I think that having `#[cfg(feature = deterministic)]` could be interesting.
Right now, I don't have anything else that stops me besides avoiding floats.

On the other hand, I guess that different Blockchain projects will have differentiations regarding what they would like to have. I'm afraid this feature might be a bit immature right now... (IMHO)

Personally, I love the freedom to enable/disable each feature in isolation as it's right now (and there aren't many flags right now).

Thank you!

YaronWittenstein:
Hi @yurydelendik @sunfishcode 

I'll need this feature pretty soon, and I'd rather not work with a forked version.
Please let me know if there's anything I can help with, in order to push this feature/similar forward.

Thank you!

yurydelendik (changes requested):
Let's do this while we thinking about effect on the code base and future of the deterministic instructions subset:

- Add `deterministic` feature, and disable it by default
- Change `enable_floats` to `deterministic_only` and move it inside `cfg()`
   * P.S. feature is disabled though default can be assigned to `true`
- Change `check_floats_enabled` to `check_non_deterministic_enabled` (and add `cfg()` there too)
- Make tests available with deterministic feature only, maybe add `cargo test --features deterministic ..` to the test-all.sh

sunfishcode:
Another feature with nondeterminism is shared memory.

YaronWittenstein:
Hi @yurydelendik 
I've made changes according to your guidance
Thank you!

YaronWittenstein:
@yurydelendik CI failed, I'm checking why.

@sunfishcode I'll need to add the `deterministic` feature to `check_shared_memarg_wo_align` too right?

sunfishcode:
@YaronWittenstein I believe `check_shared_memarg_wo_align` is only called when `check_threads_enabled` returns true, so it shouldn't need an extra check.

YaronWittenstein:
tests pass now :)

src/operators_validator.rs
```diff
@@ -488,6 +494,20 @@ impl OperatorValidator {
         Ok(())
     }
 
+    #[cfg(feature = "deterministic")]
+    fn check_non_deterministic_enabled(&self) -> OperatorValidatorResult<()> {
+        if !self.config.deterministic_only {
+            return Err("floats support is not enabled");
```
yurydelendik: nit: fix error message

yurydelendik (approved):
looks good with error message fix.

I wonder if we need explicitly enable/disable shared memory access via this feature as well. But we can iterate over this later.

src/operators_validator.rs
YaronWittenstein: fixed

YaronWittenstein: Does this change require a version bump?

yurydelendik: it does -- I'll do it

yurydelendik:
Thank you for the patch.

YaronWittenstein:
Thank you!

PR #125, opened 2019-08-19
wabt testing and mut globals fix

yurydelendik:


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "spec"]
+	path = spec
+	url = https://github.com/WebAssembly/spec
```
sunfishcode: Another option is to use `https://github.com/WebAssembly/testsuite`, which is a mirror of just the `test/core` directory of the `spec` repo, though I don't have a strong opinion about which you use here.

sunfishcode (approved):
It's not clear whether anything is testing `reloc.wasm` after this PR.

Otherwise, this looks good.

.gitmodules
yurydelendik: reloc.wasm was an object file with const STACKTOP and had set_global -- pre-mutable globals spec file. I changed STACKTOP to be mutable.

yurydelendik: Okay, switched to WebAssembly/testsuite

PR #132, opened 2019-09-19
module reader: provide SectionContent enum with readers created

pchickey:
This makes it possible to use a simpler interface as used in cranelift-wasm, and elsewhere. 

I was about to re-implement this big reader/switch from `cranelift_wasm::translate_module` in my own code (with calls to different code with each section reader). Rather than do so I thought I'd see if the library can just provide it. https://github.com/cranestation/cranelift/blob/master/cranelift-wasm/src/module_translator.rs


yurydelendik:
Thank you for the patch.

PR #133, opened 2019-09-20
validator::{validate, validate_function_body} return Result instead of bool

pchickey:
This is a breaking change to the API - instead of returning a bool, the validator functions return a Result. This allows us to report the reason for & offset of the validation failure to the user.

I figure that breaking is reasonable because this exposes more information than before, and it will be pretty trivial for users to fix.

Switching to using a Result allows using the `?` operator in a bunch of places, for increased readability.

Otherwise, I just had to pass the offset into `validate_function_body` and use the operator iterator that exposes offsets.


yurydelendik:
Agree. I was planning to do the similar change. Thank you.

PR #134, opened 2019-09-20
SectionContent enum abstraction compatible with cranelift-wasm

pchickey:
Thanks for reviewing and merging #132 so quickly yesterday! Unfortunately, I was moving too fast and didn't actually try porting it to cranelift-wasm. In #132, the SectionContent I wrote had some
problems:
* an unknown custom section's name was dropped
* a binary reader was not available for any of the recognized custom
  sections

cranelift-wasm defers decoding custom sections, except for the name
section, to the `environ` by passing the raw bytes.

This patch changes all custom sections to fall under a single variant
of SectionContent. It contains 1. the name, 2. a binary reader of the
content, and 3. an Option<CustomSectionContent<'a>>. The
CustomSectionContent enum has the specialized reader for each recognized
custom section.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmparser"
-version = "0.39.0"
+version = "0.39.1"
 authors = ["Yury Delendik <ydelendik@mozilla.com>"]
 license = "Apache-2.0 WITH LLVM-exception"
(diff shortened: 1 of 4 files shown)


pchickey:
Used correctly by cranelift-wasm here: https://github.com/CraneStation/cranelift/pull/1055

yurydelendik:
Thanks

PR #135, opened 2019-09-26
Allow `[t*] -> [t*]` if blocks with no else

fitzgen:
As long as the start and end types are the same, then we don't need to require an `else` block because a no-op will trivially implements that type signature.

This also enables the multi-value proposal spec tests, which test for this.


yurydelendik:
Thank you for the patch.

PR #137, opened 2019-11-12
Remove `std` and `core` features

alexcrichton:
This commit performs a similar transformation as
bytecodealliance/wasmtime#554, with similar motivation listed out there
as well.


yurydelendik:
Two questions:

- is there/can there be users of wasmparser with no_std feature? if yes, will we break them? ( probably major version needs to bumped)
- Is there a need to remove "test-no_std.sh"?

alexcrichton:
Yeah this'll be a breaking change at least due to the removal of the features. I didn't notice the test script, but I can remove as well! That being said it's probably best to hold off for https://github.com/bytecodealliance/wasmtime/pull/554 first.

XAMPPRocky:
@alexcrichton Now that wasmtime has now removed the std features, can this PR be merged or is there more work needed?

yurydelendik:
@XAMPPRocky the actual goal of the wasmparser was to be used in no_std environment, e.g. use no memory allocations. It might be a good idea keep supporting this feature, and remove/configure/pref anything that needs allocations.

XAMPPRocky:
@yurydelendik I don't have any particular preference one way or the other on the actual change itself, though I find the reasons for the changes mentioned in bytecodealliance/wasmtime#554 reasonable. It's more that I would be hesitant about making larger PRs with this PR still open since it could likely cause merge conflicts on one side or the other.

alexcrichton:
Oops yes, sorry, this should be good to go now.

yurydelendik (commented):
@alexcrichton can you remove test-no_std.sh as well?

alexcrichton:
Sure thing, done!

yurydelendik:
Thank you

PR #141, opened 2019-11-14
Update instruction names to latest version of specs

alexcrichton:
Some names have been shuffled around in the official text format, so
reflect those changes in the instructions defined in `CamelCase` here as
well.


yurydelendik:
Thank you for the patch!

PR #142, opened 2019-11-15
Update parsing of passive element segments

alexcrichton:
* Add support for the `ref.func` instruction
* Passive segments are a sequence of operators, no longer a sequence of
  indices


src/readers/element_section.rs
```diff
@@ -109,6 +113,48 @@ impl<'a> Iterator for ElementItemsIterator<'a> {
...
+    amt: u32,
+    reader: OperatorsReader<'a>,
+}
+
+impl<'a> PassiveElementItems<'a> {
+    pub fn get_count(&self) -> u32 {
+        self.amt
+    }
+
+    pub fn get_next_func_idx<'b>(&mut self) -> Result<Option<u32>>
```
yurydelendik: It will be nice add `PassiveElementItem` enum,  `read()` + iterator for the items.

Do you need `'b` lifetime here?

alexcrichton: Oops `'b` isn't necessary yeah, would you prefer if the iterator were added in this PR? (to me it seemed a bit wordy to add and perhaps overkill for the use case)

yurydelendik: > to me it seemed a bit wordy to add and perhaps overkill for the use case

Let's keep iterator for future. Thought let's make `pub fn read(&self) -> Result<PassiveElementItem>` / `enum PassiveElementItem { Null, Func(u32) }` change here? (We have too many `u32`)

alexcrichton: Sure thig done!

yurydelendik (approved):

yurydelendik:
Thank you!

PR #143, opened 2019-11-15
Parse the `table.fill` instruction

alexcrichton:
Add support for the experimental `table.fill` instruction from its
proposal.

### DIFF
--- a/src/operators_validator.rs
+++ b/src/operators_validator.rs
@@ -1704,6 +1704,15 @@ impl OperatorValidator {
                 }
                 self.func_state.change_frame_with_type(1, Type::I32)?;
             }
+            Operator::TableFill { table } => {
+                self.check_bulk_memory_enabled()?;
+                let ty = match resources.tables().get(table as usize) {
+                    Some(ty) => ty.element_type,
+                    None => return Err("table index out of bounds"),
+                };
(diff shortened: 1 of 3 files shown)


src/operators_validator.rs
```diff
@@ -1704,6 +1704,14 @@ impl OperatorValidator {
                 }
                 self.func_state.change_frame_with_type(1, Type::I32)?;
             }
+            Operator::TableFill { table } => {
+                self.check_bulk_memory_enabled()?;
+                if table as usize >= resources.tables().len() {
+                    return Err("table index out of bounds");
+                }
+                self.check_operands(&[Type::I32, Type::AnyRef, Type::I32])?;
```
fitzgen: is `table.fill` for `anyref` only? one can't do a `table.fill` with `funcref` tables? I would expect that this would get the type from the table, and use that type in this check, rather than hardcoding `anyref`, but I also haven't looked at the spec.

fitzgen (approved):
LGTM other than the one comment below

src/operators_validator.rs
alexcrichton: Oops good point, this should indeed be the table type, updated now!

yurydelendik:
Thank you

PR #144, opened 2019-11-15
Update names of atomic instructions

alexcrichton:
Update all names to reflect the updated spec names as of now.

* Rename "wake" to "notify"
* Use an "atomic" word in the wake/wait/fench instructions
* Remove the "U" to the end of all the rmw instructions


fitzgen (approved):
:+1:

PR #145, opened 2019-11-15
Update SIMD instructions with upstream names

alexcrichton:
* Move some "S" and "U" to the end
* Switch the location of "Sat"
* Update the "LoadSplat" instructions


yurydelendik:
Thank you!

PR #146, opened 2019-11-16
Move from `wabt` to `wast` to parsing `*.wast` files

alexcrichton:
This allows us to bypass updating the `wabt` crate and go straight to the current versions of specs, allowing us to run most of the simd and reference types test suites. A number of various and miscellaneous fixes were also added along the way to get the new tests passing.


yurydelendik (approved):
Thank you

PR #150, opened 2019-11-27
Fix passive element reader

yurydelendik:
The #142 adds parsing of passive element segment. It did not provide support for `WasmDecoder` logic, which caused https://github.com/bytecodealliance/wasmtime/issues/644.

This patch:
- fixes `Parser` and `validate` to support `PassiveElementsEntry` state
- adds test/exampe for passive elements reading

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmparser"
-version = "0.42.1"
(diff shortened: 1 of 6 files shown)


alexcrichton:
Sorry for the belated review, but looks reasonable to me!

PR #151, opened 2019-11-27
Add support for multiple tables to table.copy/table.init

eqrion:
I'd like to use this with Cranelift/Spidermonkey for reftypes support.

The spec change for the order of immediates is [here](https://github.com/WebAssembly/bulk-memory-operations/issues/117)


yurydelendik:
Thank you for the patch!

Issue #148, opened 2019-11-25
Implement `Hash` on types


RReverser:
Currently even data enums and structs (e.g. `Type`, `FuncType`, `NameEntry`, etc.), are missing `Hash` implementation, which makes it hard to use them in a `HashMap` for caching.

Would it be possible to add simple `#[derive(Hash)]` to all non-builder structs?

PR #152, opened 2019-11-29
Document and slightly refactor BinaryReader

XAMPPRocky:
This PR documents all of `wasmparser::BinaryReader`'s public methods, adds derives for common std traits, as well refactors some parts to remove clippy warnings. The only change in behaviour is `read_u64` will no longer advance its position and still fail if the `BinaryReader` has more than four bytes but less than eight. This would lead to it's position changing and making it harder to find the correct starting position.

This also changes `SectionHeader` to be `pub(crate)` removing from the public API since there is no way to construct one from `BinaryReader` in the public API.

### DIFF
--- a/src/binary_reader.rs
+++ b/src/binary_reader.rs
@@ -14,6 +14,7 @@
  */
 
 use std::boxed::Box;
+use std::convert::TryInto;
 use std::str;
 use std::vec::Vec;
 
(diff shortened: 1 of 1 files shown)


src/binary_reader.rs
```diff
@@ -13,9 +13,7 @@
  * limitations under the License.
  */
 
-use std::boxed::Box;
-use std::str;
-use std::vec::Vec;
+use std::{boxed::Box, convert::TryInto, str, vec::Vec};
```
yurydelendik: not sure it makes it easier to read

src/binary_reader.rs
```diff
@@ -38,36 +36,43 @@ fn is_name_prefix(name: &str, prefix: &'static str) -> bool {
...
 
+    /// Returns a new slice between `start` and `end - 1` from `data`.
     pub fn slice<'a>(&self, data: &'a [u8]) -> &'a [u8] {
         &data[self.start..self.end]
     }
 }
 
 /// A binary reader of the WebAssembly structures and types.
-#[derive(Clone)]
+#[derive(Clone, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
yurydelendik: we don't  need`Eq, Hash, Ord, PartialEq, PartialOrd` here

src/binary_reader.rs
```diff
@@ -1617,6 +1712,7 @@ impl<'a> BrTable<'a> {
 ///     }
 /// }
 /// ```
+#[derive(Clone, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
yurydelendik: we don't really need `Eq, Hash, Ord, PartialEq, PartialOrd` here

src/binary_reader.rs
XAMPPRocky: This will change once #137 is merged anyway to just `use std::convert::TryInto;`. It's generally considered good style to use `{}` over multiple line statements nowadays.

src/binary_reader.rs
XAMPPRocky: See #148, even if the library itself isn't using these, people may want to use the structs and enums where these traits are required. There generally isn't a reason to not derive the std traits unless you have an astronomical amount that would affect compile time.

yurydelendik: I don't think #148 implies do it for iterators and readers (it's okay for other data structures mentioned there)

src/binary_reader.rs
yurydelendik: > It's generally considered good style to use {} over multiple line statements nowadays.

Mostly agree, but in this case it is trying to lump types from multiple levels of unrelated paths.

XAMPPRocky:
@yurydelendik Updated with feedback.

src/binary_reader.rs
```diff
@@ -1560,6 +1653,11 @@ impl<'a> BrTable<'a> {
         self.cnt
     }
 
+    /// Returns whether `BrTable` is empty.
```
pepyakin:
```suggestion
    /// Returns whether `BrTable` doesn't have any labels apart from the default one.
```

src/binary_reader.rs
```diff
@@ -38,36 +39,43 @@ fn is_name_prefix(name: &str, prefix: &'static str) -> bool {
     name.starts_with(prefix)
 }
 
-const WASM_MAGIC_NUMBER: u32 = 0x6d736100;
+const WASM_MAGIC_NUMBER: u32 = 0x6d73_6100;
```
pepyakin: not sure if this is the best change. Canonical signature is actually `b"\0asm"`, and now it is divided by 2 part.

If you ask me, I'd prefer to avoid reading it as LE `u32` and just specify it as a straight byte array, like `const WASM_MAGIC: [u8; 4] = b"\0asm";`.

XAMPPRocky: Works for me, I only made the change because clippy hates long numbers without seperators. I've changed it to a byte array and updated the relevant code.

src/binary_reader.rs
pepyakin: I think this one was reverted by another change

yurydelendik:
Thank you for the patch!

PR #153, opened 2019-12-02
Fix clippy warnings

XAMPPRocky:
This fixes all the default clippy warnings, except for the warning that `ValidatingParser::process_state` has a high cyclomatic complexity.


yurydelendik:
Thank you for the patch!

Issue #155, opened 2019-12-11
WasmDecoder example is invalid


Robbepop:
Hi,

today I was trying to use wasmparser guided by its example in `Parser`:

A slightly modified version of it:
```rust
fn main() {
    let wasm = fs::read("incrementer.wasm").expect("couldn't read Wasm input");
    let mut validating_parser = ValidatingParser::new(wasm.as_slice(), None);
    let mut function_readers = Vec::new();
    loop {
        match validating_parser.read() {
            ParserState::Error(_) | ParserState::EndWasm => break,
            ParserState::BeginFunctionBody { .. } => {
                function_readers.push(validating_parser.create_binary_reader());
            }
            _ => continue,
        }
    }
    for (i, reader) in function_readers.iter_mut().enumerate() {
        println!("\nFunction {}", i);
        reader.skip_bytes(3); // I had to insert this!
        while let Ok(ref op) = reader.read_operator() {
            println!("  {:?}", op);
        }
    }
}
```

However, in order to make the parsed functions show up correctly I had to do some modifications to the `function_readers` loop:

```rust
for (i, reader) in function_readers.iter_mut().enumerate() {
    println!("\nFunction {}", i);
    let mut local_count = reader.read_local_count().unwrap();
    let mut local_decls = Vec::new();
    let n = local_count;
    for _ in 0..n {
        let local_decl = reader.read_local_decl(&mut local_count).unwrap();
        local_decls.push(local_decl);
    }
    println!("    locals: {:?}", local_decls);
    println!("    local_count¹ = {}", n);
    println!("    local_count² = {}", local_count);
    while let Ok(ref op) = reader.read_operator() {
        println!("  {:?}", op);
    }
}
```

I am not sure if this application is 100% correct but at least functions are showing up correctly with this modification of the example.

yurydelendik:
Fixed by #156

PR #156, opened 2019-12-12
Improve WasmDecoder::create_binary_reader example

Robbepop:
Improves the code example for `WasmDecoder::create_binary_reader` to make the usage intent more clear according to https://github.com/bytecodealliance/wasmparser/issues/155.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1102,11 +1102,19 @@ impl<'a> WasmDecoder<'a> for Parser<'a> {
     /// #              0x80, 0x80, 0x80, 0x0, 0x0, 0x0, 0xb];
     /// use wasmparser::{WasmDecoder, Parser, ParserState};
     /// let mut parser = Parser::new(data);
+    /// let mut types = Vec::new();
+    /// let mut function_types = Vec::new();
     /// let mut function_readers = Vec::new();
     /// loop {
-    ///     match *parser.read() {
+    ///     match parser.read() {
     ///         ParserState::Error(_) |
     ///         ParserState::EndWasm => break,
+    ///         ParserState::TypeSectionEntry(ty) => {
+    ///             types.push(ty.clone());
+    ///         }
+    ///         ParserState::FunctionSectionEntry(id) => {
+    ///             function_types.push(id.clone());
+    ///         }
     ///         ParserState::BeginFunctionBody {..} => {
     ///             let reader = parser.create_binary_reader();
     ///             function_readers.push(reader);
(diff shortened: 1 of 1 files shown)


src/parser.rs
```diff
@@ -1115,9 +1123,23 @@ impl<'a> WasmDecoder<'a> for Parser<'a> {
     ///     }
     /// }
     /// for (i, reader) in function_readers.iter_mut().enumerate() {
-    ///     println!("Function {}", i);
+    ///     // Access the function type through the types table.
+    ///     let ty = &types[function_types[i] as usize];
+    ///     println!("\nFunction {} of type {:?}", i, ty);
```
yurydelendik, line 1128: I don't think that this change (and related one above) is needed to address #155.

src/parser.rs
```diff
@@ -1115,9 +1123,23 @@ impl<'a> WasmDecoder<'a> for Parser<'a> {
...
+    ///     // Read the local declarations required by the function body.
+    ///     let mut local_count = reader.read_local_count().unwrap();
+    ///     let mut local_decls = Vec::new();
+    ///     let n = local_count;
```
yurydelendik: at this point `n == reader.read_local_count().unwrap()` and `local_count == 0`

src/parser.rs
```diff
@@ -1115,9 +1123,23 @@ impl<'a> WasmDecoder<'a> for Parser<'a> {
...
+    ///     for _ in 0..n {
+    ///         let local_decl = reader.read_local_decl(&mut local_count).unwrap();
+    ///         local_decls.push(local_decl);
+    ///     }
+    ///     println!("    locals: {:?}", local_decls);
+    ///     println!("    local_count (before) = {}", n);
```
yurydelendik: No need for `n`

yurydelendik (approved):
Looks good with changes.

I mention that using WasmDecoder is not effective way to use the library. Use `ModuleReader` and friends instead.

Robbepop:
> Looks good with changes.
> 
> I mention that using WasmDecoder is not effective way to use the library. Use `ModuleReader` and friends instead.

Thanks for the review.
Hmm, ... I was reading through the docs of `wasmparser` crate and haven't seen a single quote for the preference of `ModuleReader`. Also the examples in the `README` is not using it but the `Parser` instead. Are there tutorials/guides/examples that I have missed somewhere?

src/parser.rs
Robbepop: Yeah but it is nice for users to see how they can query the function type etc.
For any kind of serious introspection you need this information anyways.
I will remove those additional `println` statements though.

src/parser.rs
Robbepop: But `n` is shown to me (in the example for my `wasm` input) as `1`.
The `local_count (after)` is shown as `4`. (after the loop)

src/parser.rs
Robbepop: This prints:
```
local_count (before) = 1
local_count (after) = 4
```
For my `wasm` input.
So for me it was a nice info to understand what is happening since I wasn't able to infere that from the given docs.
But I will remove those additional `println` statements and `n` because you are right in that it shows more than needed for the example.

Robbepop:
@yurydelendik this is ready for merge if you are okay now

src/parser.rs
```diff
@@ -1115,9 +1123,19 @@ impl<'a> WasmDecoder<'a> for Parser<'a> {
...
     /// }
     /// for (i, reader) in function_readers.iter_mut().enumerate() {
-    ///     println!("Function {}", i);
+    ///     // Access the function type through the types table.
+    ///     let ty = &types[function_types[i] as usize];
+    ///     println!("\nFunction {} of type {:?}", i, ty);
+    ///     // Read the local declarations required by the function body.
+    ///     let mut local_count = reader.read_local_count().unwrap();
+    ///     let mut local_decls = Vec::new();
+    ///     for _ in 0..local_count {
```
yurydelendik:
```
    ///     let local_decls_count = reader.read_local_count().unwrap();
    ///     let mut local_decls = Vec::new();
    ///     let mut local_count = 0;
    ///     for _ in 0..local_decls_count {
```

src/parser.rs
yurydelendik: The reading of locals comes directly from wasm binary encoding, see https://webassembly.github.io/spec/core/binary/modules.html#binary-local

yurydelendik:
> Also the examples in the `README` is not using it but the `Parser` instead. Are there tutorials/guides/examples that I have missed somewhere?

Examples can be found in the doc comments. The README is missing the recommended API example indeed.

yurydelendik:
Thank you for the patch.

PR #158, opened 2019-12-16
Add new operations

abrown:
Recently the [SIMD spec] has added many new operations. This PR attempts to bring wasmparser up to speed with those new operations. @yurydelendik, I was not very confident about what to do about the alignment of the `Load*x*` operations--could you take a look?

[SIMD spec]: https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md


abrown:
For reference, I am using the opcodes from the [SIMD binary encoding] page.

[SIMD binary encoding]: https://github.com/WebAssembly/simd/blob/master/proposals/simd/BinarySIMD.md

src/operators_validator.rs
```diff
@@ -1337,7 +1337,13 @@ impl OperatorValidator {
                 }
                 self.func_state.change_frame_with_type(0, Type::AnyFunc)?;
             }
-            Operator::V128Load { ref memarg } => {
+            Operator::V128Load { ref memarg }
+            | Operator::I16x8Load8x8S { ref memarg }
```
yurydelendik: Yeah, the alignment needs to match the data alignment we are trying to load, e.g. 0 for `I16x8Load8x8S`? (I did similar fix at https://github.com/bytecodealliance/wasmparser/pull/140/commits/00821ddfd1659e282d619bfb0676101c9732679a#diff-3b23c7d600fea8d9864bdba37492999d)

abrown: I think actually `I16x8Load8x8S` should have a `max_align` of 3 if I read the [spec] right:
```
2^memarg.align ≤ N/8​
2^3 ≤ 64/8​
```


[spec]: https://webassembly.github.io/spec/core/bikeshed/index.html#-tmathsfhrefsyntax-instr-memorymathsfloadnmathsf_hrefsyntax-sxmathitsxhrefsyntax-memargmathitmemarg

yurydelendik: trying to apply  e.g.`i32.load8_s` validation logic here: so for `I16x8Load8x8S`, `t` = I16(x8) and `N` = 8, which makes `2^memarg.align ≤ 1` or `memarg.align ≤ 0`

abrown: Doesn't `N = 8x8 = 64`?

abrown:
Also, when I run the spec tests again I see:

```
Validation error: module did not validate: stack size does not match block type (at offset 630)
```

It seems like there are other changes necessary as well?

yurydelendik (approved):
Thank you for the patch.

There is a question about SIMD memarg alignment, but it can be addressed later.

PR #166, opened 2019-12-18
Update parsing of passive elements/simd memargs

alexcrichton:
It looks like most of the SIMD instructions which take a `MemArg` don't
need to assert that they're a specific alignment, instead they allow
many alignments (exercised in spec tests).

Additionally the encoding of passive element sections has pretty
radically changed from the original implementation. This commit brings
`wasmparser` up to date with the current version of the spec. The
notable changes here are that the flags field is a bitflags field which
has a bunch of bits indicating various types of payloads that may or may
not be present.


alexcrichton:
ping @yurydelendik, wanted to make sure this didn't fall under the radar!

src/parser.rs
```diff
@@ -117,12 +117,12 @@ pub enum ParserState<'a> {
     EndFunctionBody,
     SkippingFunctionBody,
 
-    BeginActiveElementSectionEntry(u32),
-    ElementSectionEntryBody(Box<[u32]>),
-    PassiveElementSectionEntry {
+    BeginElementSectionEntry {
+        /// `None` means this is a passive or defined entry
+        table: Option<u32>,
```
yurydelendik: Add enum of Passive, Defined, Active(u32) instead of Option.

yurydelendik (commented):
Small change for `table: Option<u32>`. I would like to find spec for SIMD memory alignment change (perhaps separate PR for that as well).

Do you have any tests to verify bulk memory changes?

alexcrichton:
Updated!

For the the spec changes/tests I get the impression that all of this is still pretty heavily in flux, so most of this isn't strictly codified but rather floating around in people's heads until it's actually codified. I've tested these changes in that I've updated [`wasmprinter`](https://github.com/bytecodealliance/wasmprinter/pull/5) to use this and it agrees with `wabt` on all encodings (at the time that was created).

The SIMD changes all mention a memarg [in their text spec](https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#load-and-splat) but only some mention it in the [binary spec](https://github.com/WebAssembly/simd/blob/master/proposals/simd/BinarySIMD.md). The tests, however, have [all sorts of alignments](https://github.com/WebAssembly/simd/blob/52c76230c9473f5648400bd9401c6868d892345c/test/core/simd/simd_load_splat.wast#L26-L30) throughout them which leads me to believe that the alignment isn't strictly required to be one particular value.

yurydelendik:
> The tests, however, have all sorts of alignments throughout them which leads me to believe that the alignment isn't strictly required to be one particular value.

That is what throws me off. Let's remove SIMD change from this PR to not delay landing of parsing elements. There is also https://github.com/WebAssembly/simd/issues/162 and #170

yurydelendik (approved):
Looks good without SIMD changes. Do you know if we can enable tests from testsuite?

alexcrichton:
Ok! I've reverted the SIMD changes now. I think a number of tests are already running and the updates to the `wast` crate means we're already reading the new encodings.

yurydelendik:
Thank you for the patch!

PR #168, opened 2020-01-06
Improve support for reftypes

eqrion:
There's been a couple of recent changes to the reference types proposal that are required for CL/Spidermonkey.


yurydelendik:
Thank you for the patch!

Are there are tests in https://github.com/WebAssembly/testsuite we can enable?

eqrion:
> Thank you for the patch!
> 
> Are there are tests in https://github.com/WebAssembly/testsuite we can enable?

Good question. It looks like most reference-types tests are currently passing. I tested out the ones still being skipped and found that two are now passing. I'll put up a PR for them.

PR #169, opened 2020-01-07
Make memarg alignment error message match expected message in SIMD spec tests

abrown:
See expected error message in https://github.com/WebAssembly/testsuite/blob/814fd73e25efbe96128e46944749ee192d0165fd/proposals/simd/simd_align.wast#L65.

### DIFF
--- a/src/operators_validator.rs
+++ b/src/operators_validator.rs
@@ -497,7 +497,7 @@ impl OperatorValidator {
         self.check_memory_index(0, resources)?;
         let align = memarg.flags;
         if align > max_align {
-            return Err("align is required to be at most the number of accessed bytes");
+            return Err("alignment must not be larger than natural");
         }
         Ok(())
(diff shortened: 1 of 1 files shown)


abrown:
@yurydelendik I observed the following when running the SIMD spec tests in wasmtime:

```
for directive on tests/spec_testsuite/proposals/simd/simd_align.wast:60:1
assert_invalid: expected alignment must not be larger than natural, got align is required to be at most the number of accessed bytes (at offset 30)
```

But now I can't seem to make that test fail any more (not sure if it is due to Cargo.lock, cached builds, etc.). I'm going to leave this as WIP until I can replicate or you think this is the right change to make (i.e. should the `Err` messages from `wasmparser` match the `assert_invalid` strings?).

yurydelendik:
I think it is a right change

PR #170, opened 2020-01-07
Allow memory alignment hints to be within [0, max_align]

abrown:
As discussed in https://github.com/WebAssembly/simd/issues/162, memory alignment hints allow the full range of values between 0 (1-byte alignment) and max_align (`log2(source byte size)`). For example, an 8-byte (64-bit) load should allow `memarg.align` values of [0, 1, 2, 3] (see https://github.com/WebAssembly/simd/issues/162#issuecomment-568064747). 

This change also attempts to make the comparison a bit more clear.

### DIFF
--- a/src/binary_reader.rs
+++ b/src/binary_reader.rs
@@ -706,9 +706,9 @@ impl<'a> BinaryReader<'a> {
         })
     }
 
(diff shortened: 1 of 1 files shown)


yurydelendik (approved):
Looks good (w/tests passing)

yurydelendik:
Thank you for the patch

PR #172, opened 2020-01-12
Generalize Wasm Validation

Robbepop:
This PR extends the `WasmModuleResources` trait in order to make it possible for 3rd party crates to use their own types while reusing the validation infrastructure provided by the `wasmparser` crate.

The core of this change is reflected by the changes to the `WasmModuleResources` trait:

## Overview

These are the new validation traits introduced by this PR:

2020-01-16-122736_1320x241_scrot

## Old Interface

- Uses slices
- Uses `wasmparser` types directly

```rust
pub trait WasmModuleResources {
    fn types(&self) -> &[FuncType];
    fn tables(&self) -> &[TableType];
    fn memories(&self) -> &[MemoryType];
    fn globals(&self) -> &[GlobalType];
    fn func_type_indices(&self) -> &[u32];
    fn element_count(&self) -> u32;
    fn data_count(&self) -> u32;
}
```

## New Interface

- Associated type allow customization points
- No longer depends on slices as interfacing type
- We could even go further with iterators but I wanted to implement this as a practical next step (incremental)

```rust
pub trait WasmModuleResources {
    /// The function type used for validation.
    type FuncType: WasmFuncType;
    /// The table type used for validation.
    type TableType: WasmTableType;
    /// The memory type used for validation.
    type MemoryType: WasmMemoryType;
    /// The global type used for validation.
    type GlobalType: WasmGlobalType;

    /// Returns the type at given index.
    fn type_at(&self, at: u32) -> Option<&Self::FuncType>;
    /// Returns the table at given index if any.
    fn table_at(&self, at: u32) -> Option<&Self::TableType>;
    /// Returns the linear memory at given index.
    fn memory_at(&self, at: u32) -> Option<&Self::MemoryType>;
    /// Returns the global variable at given index.
    fn global_at(&self, at: u32) -> Option<&Self::GlobalType>;
    /// Returns the function signature ID at given index.
    fn func_type_id_at(&self, at: u32) -> Option<u32>;

    /// Returns the number of elements.
    fn element_count(&self) -> u32;
    /// Returns the number of bytes in the Wasm data section.
    fn data_count(&self) -> u32;
}
```

## Benchmarks

tl;dr: No change in performance detected.

`master`:

```
     Running target/release/deps/benchmark-de6568819adee6d2
it works benchmark      time:   [1.7750 us 1.8170 us 1.8710 us]                               
                        change: [-9.1930% -2.5691% +3.5327%] (p = 0.47 > 0.05)
                        No change in performance detected.
Found 12 outliers among 100 measurements (12.00%)
  4 (4.00%) high mild
  8 (8.00%) high severe

validator no fails benchmark                                                                             
                        time:   [5.5755 us 5.7014 us 5.8478 us]
                        change: [+0.2051% +3.4391% +7.0131%] (p = 0.06 > 0.05)
                        No change in performance detected.
Found 13 outliers among 100 measurements (13.00%)
  5 (5.00%) high mild
  8 (8.00%) high severe

validate benchmark      time:   [5.5023 us 5.6307 us 5.7774 us]                                
                        change: [-2.1047% +0.6309% +3.8226%] (p = 0.70 > 0.05)
                        No change in performance detected.
Found 14 outliers among 100 measurements (14.00%)
  3 (3.00%) high mild
  11 (11.00%) high severe
```

`generalize-validation`:

```
     Running target/release/deps/benchmark-de6568819adee6d2
it works benchmark      time:   [1.7702 us 1.8144 us 1.8714 us]                               
                        change: [-5.6447% -2.2495% +1.0744%] (p = 0.21 > 0.05)
                        No change in performance detected.
Found 15 outliers among 100 measurements (15.00%)
  2 (2.00%) high mild
  13 (13.00%) high severe

validator no fails benchmark                                                                             
                        time:   [5.8344 us 6.0047 us 6.2036 us]
                        change: [-0.4388% +3.6992% +8.1534%] (p = 0.08 > 0.05)
                        No change in performance detected.
Found 18 outliers among 100 measurements (18.00%)
  4 (4.00%) high mild
  14 (14.00%) high severe

validate benchmark      time:   [5.4699 us 5.5826 us 5.7223 us]                                
                        change: [-3.7465% -0.3642% +3.1391%] (p = 0.84 > 0.05)
                        No change in performance detected.
Found 14 outliers among 100 measurements (14.00%)
  5 (5.00%) high mild
  9 (9.00%) high severe
```

## Future Steps

This PR introduced the `WasmType` trait that has a compatibility method allowing user code to convert their custom Wasm types into Wasm type the `wasmparser` crate understands. This was needed because the `wasmparser` crate stores those types into some data structures during validation.

For a future PR we might be able to remove this compatibility method and make the whole validation generic over some user defined custom Wasm types. However, I am not sure if the changes required for this are actually worth the efforts which is why I did not include those changes into this PR.

The `WasmType` trait and its compatibility method `to_parser_type`:

```rust
/// Types that quality as Wasm types for validation purposes.
///
/// Must be comparable with `wasmparser` given Wasm types and
/// must be comparable to themselves.
pub trait WasmType: PartialEq<crate::Type> + PartialEq + Eq {
    /// Converts the custom type into a `wasmparser` known type.
    fn to_parser_type(&self) -> crate::Type;
}
```


sunfishcode:
Interesting! One thing to keep in mind here is that an alternative to this approach would be to have third-party users just create their own `ValidatingParserResources` (wasmparser could make its version `pub` if it'd be useful) by copying data from their own data structures into it. That obviously has some overhead, but it's only handful of dynamic allocations and likely not that much total memory compared to the size of a corresponding wasm module.

A big question about the patch is, what is the performance impact of this patch? First, we have a some benchmarks in the repository that can be run with `cargo bench`, which we should make sure don't regress. The tests used there are all small wasm modules, so it'd also be good to measure the performance on some larger wasm modules as well.

Also, we need to be careful about dynamic allocations, which can cause performance effects that are sometimes harder to observe outside of larger applications. At a first pass, it appears there are a few new dynamic allocations here, in `wasm_func_type_inputs` and `wasm_func_type_outputs`. How difficult would it be to avoid those?

Robbepop:
Thank you for the comments.

> One thing to keep in mind here is that an alternative to this approach would be to have third-party users just create their own ValidatingParserResources (wasmparser could make its version pub if it'd be useful) by copying data from their own data structures into it. That obviously has some overhead, but it's only handful of dynamic allocations and likely not that much total memory compared to the size of a corresponding wasm module.

I'd say that the approach taken by this PR has a lot lower overhead than the alternative you describe here since it won't require to copy data around nor would it enforce dynamic allocations.

> A big question about the patch is, what is the performance impact of this patch? First, we have a some benchmarks in the repository that can be run with cargo bench, which we should make sure don't regress. The tests used there are all small wasm modules, so it'd also be good to measure the performance on some larger wasm modules as well.

That's neat! I will definitely benchmark it against the current `master` and report back the outcome.

> Also, we need to be careful about dynamic allocations, which can cause performance effects that are sometimes harder to observe outside of larger applications. At a first pass, it appears there are a few new dynamic allocations here, in wasm_func_type_inputs and wasm_func_type_outputs. How difficult would it be to avoid those?

It would be pretty trivial to remove those allocations you have found and I think I have even written a comment about this. I was just too lazy to do it but if it is important to you I can update this PR with the changes needed.

Robbepop:
Benchmark Results

`cargo bench` on `master` branch:

```
     Running target/release/deps/benchmark-de6568819adee6d2
it works benchmark      time:   [1.7518 us 1.7634 us 1.7789 us]                               
                        change: [-5.8538% -2.2736% +0.7621%] (p = 0.21 > 0.05)
                        No change in performance detected.
Found 17 outliers among 100 measurements (17.00%)
  3 (3.00%) low mild
  6 (6.00%) high mild
  8 (8.00%) high severe

validator no fails benchmark                                                                             
                        time:   [5.5400 us 5.6826 us 5.8534 us]
                        change: [-6.1375% -1.6758% +2.8898%] (p = 0.48 > 0.05)
                        No change in performance detected.
Found 19 outliers among 100 measurements (19.00%)
  5 (5.00%) high mild
  14 (14.00%) high severe

validate benchmark      time:   [5.4257 us 5.4568 us 5.4991 us]                                
                        change: [-4.9285% -1.9782% +0.5347%] (p = 0.18 > 0.05)
                        No change in performance detected.
Found 16 outliers among 100 measurements (16.00%)
  6 (6.00%) high mild
  10 (10.00%) high severe
```

`cargo bench` on `generalize-validation` branch:

```
     Running target/release/deps/benchmark-de6568819adee6d2
it works benchmark      time:   [1.8327 us 1.9160 us 2.0144 us]                               
                        change: [+1.9927% +6.3002% +11.238%] (p = 0.01 < 0.05)
                        Performance has regressed.
Found 22 outliers among 100 measurements (22.00%)
  1 (1.00%) high mild
  21 (21.00%) high severe

validator no fails benchmark                                                                             
                        time:   [5.9368 us 6.0219 us 6.1267 us]
                        change: [+2.6166% +6.2539% +9.6923%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 19 outliers among 100 measurements (19.00%)
  4 (4.00%) high mild
  15 (15.00%) high severe

validate benchmark      time:   [5.5948 us 5.7068 us 5.8571 us]                                
                        change: [+2.7218% +5.8568% +9.1361%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 13 outliers among 100 measurements (13.00%)
  4 (4.00%) high mild
  9 (9.00%) high severe
```

So it looks like the performance regression exists at ~5-6%.
Also note that this benchmarks the dynamically allocating `wasm_func_type_inputs` and `wasm_func_type_outputs` so maybe we can get better results once that has been fixed.

Robbepop:
I just removed the dynamic allocations of the `wasm_func_type_{inputs,outputs}` functions and can see our expected performance improvements:

```
     Running target/release/deps/benchmark-de6568819adee6d2
it works benchmark      time:   [1.7781 us 1.7956 us 1.8187 us]                               
                        change: [-8.4591% -3.6018% +1.3921%] (p = 0.16 > 0.05)
                        No change in performance detected.
Found 17 outliers among 100 measurements (17.00%)
  2 (2.00%) high mild
  15 (15.00%) high severe

validator no fails benchmark                                                                             
                        time:   [5.9056 us 6.1645 us 6.4688 us]
                        change: [+0.9319% +4.7262% +8.9154%] (p = 0.02 < 0.05)
                        Change within noise threshold.
Found 17 outliers among 100 measurements (17.00%)
  5 (5.00%) high mild
  12 (12.00%) high severe

validate benchmark      time:   [5.4440 us 5.5382 us 5.6632 us]                                
                        change: [-14.655% -8.1688% -3.0888%] (p = 0.01 < 0.05)
                        Performance has improved.
Found 15 outliers among 100 measurements (15.00%)
  9 (9.00%) high mild
  6 (6.00%) high severe
```

So in summary:

- it works: 3.6% faster
- validator: 4.7% slower
- validate: 8.1% faster

So we can even vaguely say that this patch even improves upon performance. :)

Robbepop:
We might want to try out simplifying the trait-based interface a little bit. For example I want to test if changing all those accessor methods in the main trait:

```rust
    /// Returns the type at given index.
    fn type_at(&self, at: u32) -> &Self::FuncType;
    /// Returns the table at given index.
    fn table_at(&self, at: u32) -> &Self::TableType;
    /// Returns the table at given index if any.
    fn table_at_checked(&self, at: u32) -> Option<&Self::TableType>;
    /// Returns the linear memory at given index.
    fn memory_at(&self, at: u32) -> &Self::MemoryType;
    /// Returns the global variable at given index.
    fn global_at(&self, at: u32) -> &Self::GlobalType;
    /// Returns the function signature ID at given index.
    fn func_type_id_at(&self, at: u32) -> u32;
```

to always return `Option<T>` instead of implied panicking.
Since none of these are accessed within a tight loop I guess the performance penalties of always checking are neglectible. Also we could unify `table_at` and `table_at_checked`.

```rust
    /// Returns the type at given index if any.
    fn type_at(&self, at: u32) -> Option<&Self::FuncType>;
    /// Returns the table at given index if any.
    fn table_at(&self, at: u32) -> Option<&Self::TableType>;
    /// Returns the linear memory at given index if any.
    fn memory_at(&self, at: u32) -> Option<&Self::MemoryType>;
    /// Returns the global variable at given index if any.
    fn global_at(&self, at: u32) -> Option<&Self::GlobalType>;
    /// Returns the function signature ID at given index if any.
    fn func_type_id_at(&self, at: u32) -> Option<u32>;
```

- We might consider using `usize` as index type instead of `u32`. I chose `u32` because it practically is a more constrained type which we can easily offer to the users. However, `u32` as index type looks odd to users at first.
- Also I am thinking about renaming `element_count` and `data_count` to `len_elements` and `len_data` to make them fit more into the big picture of the trait interface.

Robbepop:
I minimized the required validation `WasmModuleResources` trait interface to the following and found some potentially dangerous out-of-bounds array accesses. I have annotated these with code comments in my changes. The commits are organized as mini-steps towards this refactoring for better reviewability.

I also included a big Wasm file into the `tests` folder of the benchmarks.

```rust
pub trait WasmModuleResources {
    /// The function type used for validation.
    type FuncType: WasmFuncType;
    /// The table type used for validation.
    type TableType: WasmTableType;
    /// The memory type used for validation.
    type MemoryType: WasmMemoryType;
    /// The global type used for validation.
    type GlobalType: WasmGlobalType;

    /// Returns the type at given index.
    fn type_at(&self, at: u32) -> Option<&Self::FuncType>;
    /// Returns the table at given index if any.
    fn table_at(&self, at: u32) -> Option<&Self::TableType>;
    /// Returns the linear memory at given index.
    fn memory_at(&self, at: u32) -> Option<&Self::MemoryType>;
    /// Returns the global variable at given index.
    fn global_at(&self, at: u32) -> Option<&Self::GlobalType>;
    /// Returns the function signature ID at given index.
    fn func_type_id_at(&self, at: u32) -> Option<u32>;

    /// Returns the number of elements.
    fn element_count(&self) -> u32;
    /// Returns the number of bytes in the Wasm data section.
    fn data_count(&self) -> u32;
}
```

## Benchmarks of the latest changes

`master`:

```
     Running target/release/deps/benchmark-de6568819adee6d2
it works benchmark      time:   [1.7750 us 1.8170 us 1.8710 us]                               
                        change: [-9.1930% -2.5691% +3.5327%] (p = 0.47 > 0.05)
                        No change in performance detected.
Found 12 outliers among 100 measurements (12.00%)
  4 (4.00%) high mild
  8 (8.00%) high severe

validator no fails benchmark                                                                             
                        time:   [5.5755 us 5.7014 us 5.8478 us]
                        change: [+0.2051% +3.4391% +7.0131%] (p = 0.06 > 0.05)
                        No change in performance detected.
Found 13 outliers among 100 measurements (13.00%)
  5 (5.00%) high mild
  8 (8.00%) high severe

validate benchmark      time:   [5.5023 us 5.6307 us 5.7774 us]                                
                        change: [-2.1047% +0.6309% +3.8226%] (p = 0.70 > 0.05)
                        No change in performance detected.
Found 14 outliers among 100 measurements (14.00%)
  3 (3.00%) high mild
  11 (11.00%) high severe
```

`generalize-validation`:

```
     Running target/release/deps/benchmark-de6568819adee6d2
it works benchmark      time:   [1.7702 us 1.8144 us 1.8714 us]                               
                        change: [-5.6447% -2.2495% +1.0744%] (p = 0.21 > 0.05)
                        No change in performance detected.
Found 15 outliers among 100 measurements (15.00%)
  2 (2.00%) high mild
  13 (13.00%) high severe

validator no fails benchmark                                                                             
                        time:   [5.8344 us 6.0047 us 6.2036 us]
                        change: [-0.4388% +3.6992% +8.1534%] (p = 0.08 > 0.05)
                        No change in performance detected.
Found 18 outliers among 100 measurements (18.00%)
  4 (4.00%) high mild
  14 (14.00%) high severe

validate benchmark      time:   [5.4699 us 5.5826 us 5.7223 us]                                
                        change: [-3.7465% -0.3642% +3.1391%] (p = 0.84 > 0.05)
                        No change in performance detected.
Found 14 outliers among 100 measurements (14.00%)
  5 (5.00%) high mild
  9 (9.00%) high severe
```

src/operators_validator.rs
```diff
@@ -253,16 +268,362 @@ enum BlockType {
     If,
 }
 
+/// Types that quality as Wasm types for validation purposes.
```
yurydelendik:
```suggestion
/// Types that qualify as Wasm types for validation purposes.
```

src/operators_validator.rs
```diff
@@ -253,16 +268,362 @@ enum BlockType {
...
+fn wasm_func_type_inputs<'a, F, T>(func_type: &'a F) -> WasmFuncTypeInputs<'a, F, T>
+where
+    F: WasmFuncType<Type = T>,
+    T: WasmType + 'a,
+{
+    WasmFuncTypeInputs::new(func_type)
+}
+
+/// Returns an iterator over the output types of a Wasm function type.
+fn wasm_func_type_outputs<'a, F, T>(func_type: &'a F) -> WasmFuncTypeOutputs<'a, F, T>
```
yurydelendik: Is it possible to make `wasm_func_type_outputs` and `wasm_func_type_inputs` as methods of `WasmFuncTypeInput` perhaps outputs_iter() and inputs_iter() ?

src/operators_validator.rs
```diff
@@ -253,16 +268,362 @@ enum BlockType {
...
+    /// A type that is comparable with Wasm types.
+    type Type: WasmType;
+
+    /// Returns `true` if the global variable is mutable.
+    fn is_mutable(&self) -> bool;
+    /// Returns the content type of the global variable.
+    fn content_type(&self) -> &Self::Type;
+}
+
+/// Types  that qualify as Wasm valiation database.
```
yurydelendik:
```suggestion
/// Types that qualify as Wasm valiation database.
```

yurydelendik (commented):
Everything looks reasonable. Thank you for the benchmarking and additional comments.

Can we move newly introduce traits and primitives into separate file ("module_resources.rs"?), including `WasmModuleResources` ?

src/operators_validator.rs
Robbepop: I will look into it. :)

Robbepop:
> 
> 
> Everything looks reasonable. Thank you for the benchmarking and additional comments.
> 
> Can we move newly introduce traits and primitives into separate file ("module_resources.rs"?), including `WasmModuleResources` ?

That's a really good idea and I also thought about this.

@yurydelendik What do you think about a public crate sub-module `validation` that hosts all the validation symbols? This could potentially bring a little bit of order into the public crate interface.

This `validation` sub-module would contain:

- All the newly introduced traits.
- `WasmModuleResources` trait
- `validate` function
- `validate_function_body` function

Maybe we could also put into:

- `ValidatingOperatorParser`
- `ValidatingParser`
- `ValidatingParserConfig`
- `OperatorValidatorConfig`

Btw.: What exactly is the difference between `ValidatingParserConfig` and `OperatorValidatorConfig` ? Could we maybe unify them?

src/operators_validator.rs
Robbepop: This would be a semantic change and I am not sure if we should actually do that since then these methods would be public while at the moment they are just internal utility functions. If the method syntax is important to you I can try to enforce it somehow but might become hacky since we do not want to expose those methods publicly.

yurydelendik:
I would like to remove `WasmXXX` and `wasm_xxx` prefixes, but I can not think of a better way to group them and differentiate them from concrete types. Perhaps to move them into sub-module and dropping "wasm" prefix? Or I guess we can keep it as is for now. 

> What do you think about a public crate sub-module validation that hosts all the validation symbols? 

Let's keep this change out of this PR, we can follow up that in the separate PR or issue.

>  What exactly is the difference between ValidatingParserConfig and OperatorValidatorConfig ? Could we maybe unify them?

There was plans to perform validation only on a function, out of context of a module. So the OperatorValidatorConfig could configure only function validation. Not sure if this requirement is still the case.

Robbepop:
> I would like to remove `WasmXXX` and `wasm_xxx` prefixes, but I can not think of a better way to group them and differentiate them from concrete types. Perhaps to move them into sub-module and dropping "wasm" prefix? Or I guess we can keep it as is for now.
> 
> > What do you think about a public crate sub-module validation that hosts all the validation symbols?
> 
> Let's keep this change out of this PR, we can follow up that in the separate PR or issue.
> 
> > What exactly is the difference between ValidatingParserConfig and OperatorValidatorConfig ? Could we maybe unify them?
> 
> There was plans to perform validation only on a function, out of context of a module. So the OperatorValidatorConfig could configure only function validation. Not sure if this requirement is still the case.

Thank you very much for the response. So what is missing for merging? :)
I agree that we should not put much more work into this single PR and instead do a follow-up if needed.

yurydelendik (approved):
Looks good. I double checked with valgrind and looks like the performance did not regress. Thank you for the patch!.

PR #173, opened 2020-01-13
Enable ref-types 'binary.wast' and 'linking.wast' tests

eqrion:
These tests are now passing, possibly because of #168.


yurydelendik:
thank you for the patch

PR #175, opened 2020-01-17
Refactor BrTable

Robbepop:
Associated issue: https://github.com/bytecodealliance/wasmparser/issues/174

This PR slightly refactors the `BrTable` abstraction of the Wasm `br_table`.

- Adds a new field `default_offset` to `BrTable`
- Removes the `IntoIterator` impl for `&'a BrTable`
- Adds `targets` and `targets_and_default` iterator methods to `BrTable`
- Improves correctness of iteration of `BrTableTargets` (the iterator type)

Let me explain the changes.

- Since the default offset is a required field in the `br_table` we simply add it as a field instead of merging it with the other non-default target offsets.
- I had the feeling that the `IntoIterator` impl was confusing since it was not clear what exactly is being iterated over (with or without default offset?). So the new interface provides two different methods for those two use cases.
- The last thing is a bit more involved. In essence this changes the very simple iterator `next` implementation from
```rust
fn next(&mut self) -> Option<Self::Item> {
    self.reader.read_var_u32().ok()
}
```
to
```rust
fn next(&mut self) -> Option<Self::Item> {
    if self.start == self.end {
        return None;
    }
    let next = self
        .reader
        .read_var_u32()
        // We can expect this not to fail since we can only contruct a
        // `BrTableTargets` iterator from a proper `BrTable` instance
        // which checks the underlying buffer for being a valid `br_table`
        // entries buffer.
        .expect("expected a `br_table` target var_u32");
    self.start += 1;
    Some(next)
}
```
While this might look like a really bad change it actually makes the whole iteration more correct from what I can tell. Please prove me wrong on this because I am actually still a bit confused about what I found: The `empty-br-table` test broke after this change - note, that every other test of the whole test suite continued to work under this change.

From looking at the test snippet it seems to me that the `empty-br-table.wasm` is actually syntactically incorrect, not semantically so it should fail during parsing, not during validation but I might be wrong here.

Why is the old and super fancy small interface working? Tbh I cannot tell for sure but it obviously is more robust against too small buffers. So the iterator will silently yield less elements than there are which is what is incorrect about it imo. But I am still a bit confused because the `create_br_table` function should in theory sort out those malicious cases before they can even exist. Need to dig in deeper.

Please tell me if I am completely wrong about my assumptions so far. @yurydelendik

### DIFF
--- a/src/binary_reader.rs
+++ b/src/binary_reader.rs
@@ -367,21 +367,23 @@ impl<'a> BinaryReader<'a> {
     }
 
     fn read_br_table(&mut self) -> Result<BrTable<'a>> {
-        let targets_len = self.read_var_u32()? as usize;
-        if targets_len > MAX_WASM_BR_TABLE_SIZE {
+        let len_targets = self.read_var_u32()? as usize;
+        if len_targets > MAX_WASM_BR_TABLE_SIZE {
             return Err(BinaryReaderError {
                 message: "br_table size is out of bound",
                 offset: self.original_position() - 1,
             });
         }
         let start = self.position;
-        for _ in 0..targets_len {
+        for _ in 0..len_targets {
             self.skip_var_32()?;
         }
-        self.skip_var_32()?;
+        let default_offset = self.read_var_u32()?;
+        let end = self.position;
         Ok(BrTable {
-            buffer: &self.buffer[start..self.position],
-            cnt: targets_len as usize,
+            buffer: &self.buffer[start..end],
+            default_offset,
+            len_targets,
         })
     }
 
@@ -493,7 +495,7 @@ impl<'a> BinaryReader<'a> {
             if shift >= 25 && (byte >> (32 - shift)) != 0 {
                 // The continuation bit or unused bits are set.
                 return Err(BinaryReaderError {
-                    message: "Invalid var_u32",
+                    message: "encountered invalid var_u32 while reading",
                     offset: self.original_position() - 1,
                 });
             }
@@ -518,7 +520,7 @@ impl<'a> BinaryReader<'a> {
             }
         }
         Err(BinaryReaderError {
-            message: "Invalid var_32",
+            message: "encountered invalid var_32 while skipping",
             offset: self.original_position() - 1,
         })
     }
@@ -1694,20 +1696,24 @@ impl<'a> BinaryReader<'a> {
 }
 
 impl<'a> BrTable<'a> {
-    /// Returns the number of `br_table` entries, not including the default
-    /// label
+    /// Returns the number of `br_table` targets.
+    ///
+    /// # Note
+    ///
+    /// This does *not* include the default target.
     pub fn len(&self) -> usize {
-        self.cnt
+        self.len_targets
     }
 
-    /// Returns whether `BrTable` doesn't have any labels apart from the default one.
+    /// Returns `true` if the `br_table` has no targets apart from the default one.
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
 
-    /// Reads br_table entries.
+    /// Returns the `br_table` targets.
     ///
     /// # Examples
+    ///
     /// ```rust
     /// let buf = vec![0x0e, 0x02, 0x01, 0x02, 0x00];
     /// let mut reader = wasmparser::BinaryReader::new(&buf);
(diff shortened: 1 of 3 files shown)


yurydelendik:
> From looking at the test snippet it seems to me that the empty-br-table.wasm is actually syntactically incorrect, not semantically so it should fail during parsing, not during validation but I might be wrong here.

The test is correct. There is an issue with skip_var_32() that is skipping a number and not checking if they are in range -- that's why it happens during validation. The skip_var_xx logic needs to be fixed.

src/binary_reader.rs
```diff
@@ -367,21 +367,23 @@ impl<'a> BinaryReader<'a> {
     }
 
     fn read_br_table(&mut self) -> Result<BrTable<'a>> {
-        let targets_len = self.read_var_u32()? as usize;
-        if targets_len > MAX_WASM_BR_TABLE_SIZE {
+        let len_targets = self.read_var_u32()? as usize;
```
yurydelendik, line 370: wonder why `targets_len` renamed to `len_targets`?

src/binary_reader.rs
```diff
@@ -1732,44 +1738,87 @@ impl<'a> BrTable<'a> {
...
+pub struct BrTableTargets<'a> {
+    /// The underlying reader yielding the entries bytes.
     reader: BinaryReader<'a>,
+    /// The start byte index.
+    start: usize,
+    /// The end byte index.
+    end: usize,
 }
 
-impl<'a> IntoIterator for &'a BrTable<'a> {
```
yurydelendik, line 1755: I think we can still keep `IntoIterator` for `BrTable`: `type IntoIter = Chain<BrTableTargets, Once<u32>>`?

src/binary_reader.rs
```diff
@@ -1732,44 +1738,87 @@ impl<'a> BrTable<'a> {
...
         }
-        self.reader.read_var_u32().ok()
+        let next = self
+            .reader
+            .read_var_u32()
+            // We can expect this not to fail since we can only contruct a
+            // `BrTableTargets` iterator from a proper `BrTable` instance
+            // which checks the underlying buffer for being a valid `br_table`
+            // entries buffer.
+            .expect("expected a `br_table` target var_u32");
```
yurydelendik, line 1810: We can keep it after skip_var_32 is fixed, though it shall never be raised during testing.

src/primitives.rs
```diff
@@ -191,11 +191,15 @@ pub enum RelocType {
...
+/// A `br_table` entries representation.
 #[derive(Debug, Clone)]
 pub struct BrTable<'a> {
+    /// The underlying bytes buffer representing the already parsed targets.
     pub(crate) buffer: &'a [u8],
-    pub(crate) cnt: usize,
+    /// The number of non-default targets of the `br_table`.
+    pub(crate) len_targets: usize,
+    /// The default target offset of the `br_table`.
+    pub default_offset: u32,
```
yurydelendik, line 202: `default_offset` -> `default_target`

src/binary_reader.rs
Robbepop: We can do that. I just removed it because it is not directly clear what exactly BrTable is iterating over given `IntoIterator` whereas for types like `Vec` it is totally clear. So I thought maybe methods are a superior fit in this case and therefore replaced the `IntoIterator` impl with two methods for each explicit use case.

src/binary_reader.rs
Robbepop: tbh no good reasons. i just personally liked the name better during refactoring. will undo. ;)

src/binary_reader.rs
Robbepop: I wasn't sure if `skip_var_32` is actually broken but good to know. :) I think this implementation is actually more robust since it pointed out this potential bug and it shouldn't be a performance hit at all. But we should make sure that this is actually true. Fortunately `wasmparser` has a decent benchmark suite.

src/primitives.rs
Robbepop: ok :+1:

Robbepop (commented):
> > From looking at the test snippet it seems to me that the empty-br-table.wasm is actually syntactically incorrect, not semantically so it should fail during parsing, not during validation but I might be wrong here.
> 
> The test is correct. There is an issue with skip_var_32() that is skipping a number and not checking if they are in range -- that's why it happens during validation. The skip_var_xx logic needs to be fixed.

Thank you for letting me know. I was really confused. :sweat_smile:

Robbepop:
Closed in favor of https://github.com/bytecodealliance/wasmparser/pull/216.

Issue #167, opened 2019-12-19
Some items are missing their public exports


Robbepop:
Some items are missing their public exports which makes using them difficult.

Namely
- `ActiveElementsItemsReader`
- `ProducersFieldValueReader`

Current work around for example for `ActiveElementsItemsReader` is to use:
```rust
let mut reader = items.get_items_reader()?;
let items = {
    let mut items = Vec::new();
    while let Ok(id) = reader.read() {
        items.push(id)
    }
    items.into_boxed_slice()
};
```
... which ignores errors while reading. Instead of:
```rust
let items = items
    .get_items_reader()
    .into_iter()
    .collect::<Result<Vec<_>, _>>()?
    .into_boxed_slice();
```

I attached some pictures of the documentation that shows this effect there.

![2019-12-19-140621_1612x425_scrot](https://user-images.githubusercontent.com/8193155/71175910-edd00e80-2268-11ea-99eb-e887ccec4ff8.png)
![2019-12-19-113749_921x763_scrot](https://user-images.githubusercontent.com/8193155/71175918-f1639580-2268-11ea-86cf-5f3a36be13db.png)

PR #176, opened 2020-01-17
Add missing export for ProducersFieldValueReader

Robbepop:
Fixes issue #167 .
Due to the latest changes `ActiveElementsItemsReader` is no longer existing and thus does no longer need another public export.


yurydelendik:
thank you!

PR #180, opened 2020-01-17
Put scripts into scripts dir

Robbepop:
This PR also slightly improves the `test-all.sh` script to work better wherever it has been called from.

CI needs to be adjusted in order to work with this PR. I think I cannot do this myself.


Robbepop:
@yurydelendik Feel free to close if you don't like the idea behind the changes introduced in this PR.

yurydelendik:
>  Feel free to close if you don't like the idea behind the changes introduced in this PR.

Not all scripts are moved (e.g. "compare-master.sh") and ".github/workflows/main.yml" was not updated. It would be nice to remove "test-all.sh" and "format-all.sh" (there are old) and just use corresponding `cargo test` and `cargo fmt`. Thanks for looking into that.

PR #181, opened 2020-01-17
Update criterion 0.2 -> 0.3

Robbepop:
I thought this was going to be more work but apparently it was just a one liner.


yurydelendik:
thank you

PR #182, opened 2020-01-17
Match lane index error message expected by SIMD spec tests

abrown:
See https://github.com/WebAssembly/testsuite/blob/c70c3c8b136e5e7193135d40ec3960f4ef1cb20a/proposals/simd/simd_lane.wast#L398, e.g.


yurydelendik (approved):
Thank you!

PR #184, opened 2020-01-31
Add SIMD integer min/max

abrown:
See https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#lane-wise-integer-minimum


yurydelendik (approved):

yurydelendik:
Thank you for the patch

PR #186, opened 2020-02-04
Ensure we have correct error messages for `(assert_invalid ...)` spec tests

fitzgen:
Started on `assert_malformed` as well, but this requires deeper changes to leb128 parsing, so I figured we might as well land this part now.

### DIFF
--- a/src/operators_validator.rs
+++ b/src/operators_validator.rs
@@ -14,13 +14,11 @@
  */
 
 use std::cmp::min;
-use std::result;
-use std::str;
 
 use crate::primitives::{MemoryImmediate, Operator, SIMDLaneIndex, Type, TypeOrFuncType};
 use crate::{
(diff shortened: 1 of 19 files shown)


src/operators_validator.rs
```diff
@@ -110,7 +108,7 @@ impl FuncState {
             let len = self.stack_types.len();
             let remove_non_polymorphic = len
                 .checked_sub(last_block.stack_starts_at)
-                .ok_or("invalid block signature")?
+                .ok_or(OperatorValidatorError::new("invalid block signature"))?
```
yurydelendik: I wonder if `ok_or_else` is warranted here?

yurydelendik (approved):
Looks good. A small question about `.ok_or()`.

src/operators_validator.rs
fitzgen: yeah totally, good call

PR #191, opened 2020-02-06
Fix long running benchmarking

yurydelendik:


yurydelendik:
Still takes long time -- longer than locally.

PR #193, opened 2020-02-07
Fix dead code validation for multivalue block

### LINKED ISSUE #192: Panic on overflow with malformed multi-value module
Given this wasm file as input [foo.wasm[2].gz](https://github.com/bytecodealliance/wasmparser/files/4171056/foo.wasm.2.gz) the following will cause a panic:

```rust
fn main() {
    let input = std::fs::read("foo.wasm").unwrap();
    let config = wasmparser::ValidatingParserConfig {
        operator_config: wasmparser::OperatorValidatorConfig {
            enable_threads: false,
            enable_reference_types: false,
            enable_simd: false,
            enable_bulk_memory: false,
            enable_multi_value: true,
        },
    };
    drop(wasmparser::validate(&input, Some(config)));
}
```

with:

```
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/wasmparser`
thread 'main' panicked at 'attempt to subtract with overflow', src/operators_validator.rs:170:31
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

I tried to poke around to see if I could fix this myself, but I couldn't really understand the validation code here :(

yurydelendik:
Avoid assert/numeric overflow assotiated with "borrowing" unknown stack types from polymorphyc stack when multi-value type block is created.

Fixes #192


src/operators_validator.rs
```diff
@@ -167,15 +173,33 @@ impl FuncState {
                 return Err(OperatorValidatorError::new("stack operand type mismatch"));
             }
         }
-        let stack_starts_at = self.stack_types.len() - start_types.len();
+        let (stack_starts_at, polymorphic_values) = {
+            // When stack for last block is polymorphic, ensure that
+            // the polymorphic_values matches, and next block informed about that.
```
fitzgen: nitpick:

```suggestion
            // the polymorphic_values matches, and next block is informed about that.
```

src/operators_validator.rs
```diff
@@ -167,15 +173,33 @@ impl FuncState {
...
+            {
+                (self.stack_types.len() - start_types.len(), None)
+            } else {
+                let unknown_stack_types_len =
+                    last_block.stack_starts_at + start_types.len() - self.stack_types.len();
+                let polymorphic_values = last_block
+                    .polymorphic_values
+                    .unwrap()
+                    .max(unknown_stack_types_len);
+                last_block.polymorphic_values = Some(polymorphic_values);
```
fitzgen: This is subtle! I had to game it out with "what if we have fewer result types than start types? what if the opposite?" etc.

Nice find, fuzzers! :) and nice fix, Yury! :)

fitzgen (approved):
Looks great!

I wish that wasm2wat worked with super invalid wasm files...

```
$ wasm2wat --enable-multi-value --no-check ~/Downloads/foo.wasm 
0000081: error: invalid call function index: 2097156
```

But aha! A new contender appears:

```
$ cargo run --manifest-path ~/wasmprinter/Cargo.toml --example wasm2wat -- ~/Downloads/foo.wasm 
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `/home/fitzgen/wasmprinter/target/debug/examples/wasm2wat /home/fitzgen/Downloads/foo.wasm`
(module
  (type (;0;) (func (param i32)))
  (type (;1;) (func (param i32) (result i32)))
  (type (;2;) (func (param i32 i32)))
  (type (;3;) (func))
  (func (;0;) (type 3))
  (func (;1;) (type 0) (param i32))
  (func (;2;) (type 1) (param i32) (result i32)
    i32.const 1)
  (func (;3;) (type 1) (param i32) (result i32)
    i32.const 2)
  (func (;4;) (type 2) (param i32 i32)
    local.get 0
    unreachable
    unreachable
    unreachable
    unreachable
    unreachable
    unreachable
    unreachable
    block (param i32)  ;; label = @1
      unreachable
      drop
    end)
  (func (;5;) (type 3)
    (local i32)
    unreachable
    unreachable
    i32.mul
    i32.popcnt
    i32.div_u
    i32.sub
    i32.popcnt
    i32.div_u
    i32.sub
    local.tee 0
    global.set 0
    local.get 0
    i32.const 8
    i32.add
    i32.const 1024
    call 4
    local.get 0
    i32.const 8
    i32.add
    i32.const 25166864
    unreachable
    call 2097156
    call_indirect (type 0)
    i32.const 16
    i32.add
    global.set 0)
  (func (;6;) (type 3)
    i64.clz))
```

Would it be possible to use this wat file (potentially further reduced/simplified) as a test case instead of the wasm binary? Should be easier for anyone who has to debug what the heck is going on with it in the future.

fitzgen:
Here is a reduced WAT test case via `creduce`:

```wat
(module
  (func
    unreachable
    block (param i32)
    end))
```

yurydelendik:
> Would it be possible to use this wat file (potentially further reduced/simplified) as a test case instead of the wasm binary?

I just need to add wast testing with options/features

src/operators_validator.rs
yurydelendik: That line was not needed... actually. I found it out when was adding the wast test. Yay, tests.

fitzgen (approved):
Nice!

PR #195, opened 2020-02-18
Fix br for loop with parameters

### LINKED ISSUE #194: Multi-value validation fails for `br_if` and `loop`
The following `*.wat` file:

```wat
(module
  (func (param i32)
    (local i32)
      i32.const -64
      loop (param i32)
        local.set 0
        local.get 0
        br_if 0
      end))
```

looks to be invalid:

```
 $ wat2wasm --enable-all foo.wat
foo.wat:8:9: error: type mismatch in br_if, expected [i32] but got []
        br_if 0
        ^^^^^
foo.wat:9:7: error: type mismatch in loop, expected [] but got [i32]
      end))
      ^^^
```

but it passes wasmparser's validation currently!

yurydelendik:
Fixes #194

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmparser"
-version = "0.51.1"
+version = "0.51.2"
 authors = ["Yury Delendik <ydelendik@mozilla.com>"]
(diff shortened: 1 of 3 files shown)


yurydelendik:
trivial fix, tests included

Issue #198, opened 2020-02-24
Make wasmparser no_std again


yurydelendik:
There is still intent to support "no_std" feature for this crate. There is no need to support allocation hungry API such as `WasmParser`/`WasmDecoder`, so it will be beneficial to make this legacy API under feature -- the section readers is more performant API and has to be used at this moment. 

TODOs:
- [ ] Make `WasmParser` and friends optional via feature (and/or deprecate)
- [ ] Cleanup sections readers to not use dynamic memory `Box`, `Vec`, `HashMap`

I think that will allow us to be `no_std` without depending on `alloc` or `hashbrown` crates.

fitzgen:
What API are you imagining for things like `BrTable::read_table` where we allocate a `Box<[..]>`? Instead taking a user-provided callbacks-style trait to incrementally report the entries and let the trait implementor decide if/how to allocate?

yurydelendik:
> What API are you imagining for things like BrTable::read_table where we allocate a Box<[..]> ?

We already have BrTableIterator to read it incrementally. Is it possible improve it further to allow optimal `Iterator::collect`?

yurydelendik:
- [ ] convert validate logic into something that can be done without `Vec` (can be possible with preallocated buffers or cache-read?)

Robbepop:
While I like the proposal of going back to `no_std` this kind of conflicts with the other proposal to use `T: Read` as parser input instead of byte slices. We could work around that by creating another trait that is similar to `Read` but works in `no_std`.

PR #199, opened 2020-02-25
Support no_std

IcyDefiance:
Related to bytecodealliance/cranelift#1261.

This PR takes takes the same approach to supporting no_std that cranelift-wasm currently does. That is, it has no_std always enabled and uses either `extern crate std;` or `extern crate alloc as  std;` depending on whether or not the std feature is enabled.

It does not use no-std-compat, and it does not eliminate all use of dynamic memory as recommended in #198, therefore it does add hashbrown as an optional dependency.

To address @bnjbvr's comment [here](https://github.com/bytecodealliance/cranelift/issues/1261#issuecomment-590442143), this may not be the best way to support no_std, but considering a related crate already uses this approach, I think it's a good baseline that other approaches should be compared to.

Pros:
* can use std instead of alloc
* compiler errors if devs don't add use statements for things like Vec and String that are only implicitly available when std is enabled

Cons:
* devs must still use core instead of std, which is not idiomatic, and is part of the reason no_std support was dropped in wasmtime
  * adding a module named "std" that exports `alloc::*` and `core::*` is an easy solution, though I think it would have to be referenced with `crate::std`
  * no-std-compat effectively does the same thing, but because it's a crate, it can be referenced with just `std` if it's aliased
* requires hashbrown (and therefore two use statements instead of one) for a single use of HashSet
  * exporting hashbrown from the "std" module described above would eliminate one of those use statements
  * no-std-compat also does that
  * hashbrown is not HashDoS-resistant, and it could be confusing for users if it's exposed in the public api. I don't think either of those concerns are relevant to this library, but they're worth considering in a more general sense.

Note: I'm not sure why github says the testsuite submodule was changed. I don't think I did anything that could have affected it.


sunfishcode:
cranelift-wasm does do things this way, however as one of the people involved, no-std-compat wasn't available, and we didn't know of any other ways to do it. With the benefit of hindsight, my experience has been that trying to get everyone who works on these crates to remember what's in `core::` vs. `std::`, and explicitly declare all the `Vec`s and `String`s everywhere is a burden both for newcomers but also for folks who work on lots of different crates. It now looks like no-std-compat should make no_std support much less invasive, so I'd like to suggest we experiment with migrating wasmparser in that direction as well, and if it goes well, we can port cranelift-wasm to use it as well.

yurydelendik:
> It does not use no-std-compat, and it does not eliminate all use of dynamic memory as recommended in #198, therefore it does add hashbrown as an optional dependency.

I don't see any reason why data structures that use heap has to be present in wasmparser. Can you explain why we shall continue supporting old WasmDecoder interface?

IcyDefiance:
There were 3 reasons why I didn't try to eliminate heap allocations in this PR:
1. That's not my decision to make.
2. Since there's a desire to use this repository as a testbed for supporting no_std in other crates, there's value in learning how to do so with a hashset and the resources in alloc.
3. I think it should be a separate PR anyway, especially if it requires dropping support for anything.

That said, I think it's a good idea if it's possible. Since it's related to what I'm doing here, I might try to do that once these experiments are done.

Cargo.toml
```diff
@@ -19,10 +22,13 @@ wast = "7.0.0"
 travis-ci = { repository = "bytecodealliance/wasmparser.rs" }
 
 [features]
+default = ["std"]
 # The "deterministic" feature supports only Wasm code with "deterministic" execution
 # across any hardware. This feature is very critical for many Blockchain infrastructures
 # that rely on deterministic executions of smart contracts across different hardwares.
 deterministic = []
+std = []
+core = ["hashbrown"]
```
IcyDefiance: I learned that this is technically a breaking change for anything that uses wasmparser with `default-features = false`. That's probably worth noting, at least because it affects what version number should be changed.

yurydelendik:
> Since there's a desire to use this repository as a testbed for supporting no_std in other crates, there's value in learning how to do so with a hashset and the resources in alloc.

I would say it will be a bad decision to experiment with something that does not fit in the project's vision, and use it as a playground for it.

I'll defer the review to the @alexcrichton since he has better insights into no_std future.

alexcrichton:
I would echo @sunfishcode's concern above. The idioms in the `#![no_std]` world here are so wonky that it makes both newcomers and seasoned developers alike not really know what to do. This is unfortunately not something with a great local solution, and in my opinion it's a failing of `#![no_std]` in general.

I would personally advocate pretty strongly for a target that wants to support infallible allocation (the `alloc` crate) then it should support libstd as well, and that would make the `#![no_std]` distinction moot.

IcyDefiance:
As I understand the current situation, there are now 3 conflicting opinions:
1. bnjbvr (and whoever he's including in his use of "we") wants to use this crate to experiment with how best to support no_std
2. yurydelendik thinks eliminating alloc as part of supporting no_std is better than supporting no_std and then eliminating alloc/hashbrown
3. alexcrichton doesn't think any crate should support no_std+alloc, at least for now

Until one of these opinions becomes a decision, here's my reasoning: If this PR hasn't been closed and the "Building with no_std" section hasn't been removed from cranelift's readme, I'm going to assume that supporting no_std is still a goal, and even though it may be possible to support no_std without alloc in this repo, that won't be possible in cranelift. Between the other two opinions, bnjbvr's requires a smaller diff and still allows for eliminating heap allocations afterward, so I'll run with that.

---

For experiment 2, I've introduced no-std-compat and dropped hashbrown as a direct dependency (no-std-compat still imports it because I enabled the compat_hash feature).

Pros (compared to baseline):
* can use std instead of core
* can use std instead of a second hashbrown use statement
* no need for use statements for Vec, String, etc
* no need for the core feature

Cons:
* must add `use std::prelude::v1::*;` to access types that are normally implicitly available
  * hardly anyone is using no-std-compat right now, so hardly anyone will expect that. if this approach is used, maybe it should be explained in the readme or something?
  * it may be easier to understand if a standard is set for it to be at the top of all files. in this case, it will also require `#[allow(unused_imports)]`

Notes:
* no-std-compat 0.3.0 is broken, so I used 0.2.0. I created an issue for it [here](https://gitlab.com/jD91mZM2/no-std-compat/issues/4).
* while no-std-compat's readme recommends conditionally enabling no_std, this allows devs to forget the `use std::prelude::v1::*;` line, accidentally breaking no_std, so I made no_std unconditional

I think this is definitely an improvement. There's only one unusual thing for devs to learn, and the diff for this PR is tiny. However, it's not *quite* perfect, and I don't see a way to improve it anymore, which contributes to alexcrichton's point. The question is: is it close enough?

bnjbvr:
Fwiw, the "we" i was expressing was what seemed to be a consensus from discussing with @sunfishcode and @alexcrichton. @yurydelendik knows wasmparser better than anyone else, I think. So I'm all in favor doing what they recommend in general here, myself not being a no-std expert at all.

yurydelendik (commented):
The testsuite was updated by this PR. I think merge commit went wrong.

IcyDefiance:
Okay, I'll look at dropping alloc/hashbrown next.

And yeah, I'm not sure how I updated the testsuite, because I just did a normal merge. I'll try doing a soft reset and recommitting my changes tonight.

alexcrichton:
Again these are my own personal opinions, but I feel that even with `no-std-compat` this is not something we want to land and propagate throughout the bytecodealliance projects. I'll reiterate that "`no_std` + `alloc`" is not in the spirit of how `no_std` was designed, so it's buying into a worse developer experience from the get-go. This tries to minimize the experience, but even adding prelude imports in every module I find bizarre and enough of a deterrent to avoid this. I'm also not personally a fan of an obscure multi-line header in each crate where you have to dig real closely to figure out what's happening there. 

Another reason that I don't think that this is adequate as-is is that there is no verification. Theres' nothing stopping something from pulling in `std` itself and using it (for example in a dependency). To that end there's no way to actually check and test that we don't pull in dependencies, activate the right features, etc.

I'd reiterate again, what's the target in question here? Can the target simply be updated to compile the `std` crate?

IcyDefiance:
In my case, the target is a new OS that runs wasm programs in ring 0. It's just a toy project, but I'm not the only person with this idea: See [nebulet](https://github.com/nebulet/nebulet) and [redshirt](https://github.com/tomaka/redshirt).

Theoretically, I could probably compile a second binary with std and hack together enough of a kernel api to run it, but if this isn't merged it'll be easier for me to just maintain my own fork that has no_std support.

alexcrichton:
Er well sort of what I'm saying is that I don't think this is a hack. I'm not sure what your Rust target is, but libstd can be stubbed out to return errors for things like `thread::spawn`, you don't have to actually implement it or hook it up to kernel APIs. If the `std` crate builds for your targets then most of it will return error, and the need for `#![no_std]` is moot.

IcyDefiance:
You may have convinced me. I have more research to do before I can say that solves my use case, but I agree that no-std-compat's prelude is bizarre enough for pretty much any other reasonable solution to be preferable, so I'll give it a try.

alexcrichton:

PR #201, opened 2020-02-26
Fix typo in `ValidatingParser::current_position`

MarkMcCaskey:
Follow up to #200.  Looks like a very Dvorak-style typo, it looks like my left (where all the vowels are) and right hands were slightly out of sync!   Sorry I didn't catch the mistake earlier!


yurydelendik:
Thanks

PR #202, opened 2020-02-27
Add a `wasm-validate` example program

alexcrichton:
Shows off how to to use `validate` and makes it pretty nifty to install
from crates.io if you want.


yurydelendik (approved):

yurydelendik:
Thank you for the patch!

PR #203, opened 2020-03-05
Fix OperatorValidator for floating-point (non-deterministic) opcodes

camilbancioiu:
This pull request fixes an inverted `if` in `check_non_deterministic_enabled()`, which was allowing floating-point opcodes even when `deterministic_only` was set to `true`.

Also renames `check_non_deterministic_enabled()` to `check_deterministic_only()`, to make it clearer.

On the master branch, `Cargo.toml` defines the `deterministic` feature, which seems to be enabled during the CI tests (is it?). The test which parses the file `float_memory.0.wasm` doesn't expect an error, and fails with the fixed `OperatorValidator`. Is the test correct?


yurydelendik:
> On the master branch, Cargo.toml defines the deterministic feature, which seems to be enabled during the CI tests (is it?)

It is defined, but not enabled (as default features)

> The test which parses the file float_memory.0.wasm doesn't expect an error, and fails with the fixed OperatorValidator. Is the test correct?

The test is correct.

camilbancioiu:
@yurydelendik Thank you for your quick reply. I'll look deeper into this, it might be an issue somewhere else (we're using Wasmer, which in turn relies on Wasmparser).

alexcrichton:

PR #207, opened 2020-04-22
Renumber SIMD ops

yurydelendik:
Trying to address https://github.com/WebAssembly/simd/pull/209

Few questions:
- [x] What to do with disappeared ops: I8x16Mul, I64x2AnyTrue, I64x2AllTrue, I64x2TruncSatF64x2S, I64x2TruncSatF64x2U, F64x2ConvertI64x2S, F64x2ConvertI64x2U,
- [x] There are new ops: I32x4DotI16x8S, I8x16Abs, I16x8Abs, I32x4Abs,


abrown:
Here's the feedback I got from the meeting:
 - for old instructions that still fit in the new scheme, they were talking about keeping them around and putting them behind a flag (e.g. `--deprecated-simd-opcodes`, my own invention)
 - for old instructions that do not fit, they were removed
 - for new instructions that have not yet been standardized, we could also add them behind a flag (e.g. `--experimental-simd-opcodes`)

I'm still going back and forth on whether they should be compile-time features or runtime flags--what do you think? Runtime flags would be nice for me but we would have to wire that through wasmtime as well.

yurydelendik:
> What to do with disappeared ops: I8x16Mul, I64x2AnyTrue, I64x2AllTrue, 

Placed these three into reserved slots

> I64x2TruncSatF64x2S, I64x2TruncSatF64x2U, F64x2ConvertI64x2S, F64x2ConvertI64x2U,

Removed from encoding since they are above 0xff

> There are new ops: I32x4DotI16x8S, I8x16Abs, I16x8Abs, I32x4Abs,

These are not introduced in this PR

yurydelendik:
- [x] Tests

yurydelendik:
@alexcrichton @abrown @sunfishcode not sure about testing. I probably will disable SIMD proposal testing in this PR and when we update https://github.com/bytecodealliance/wat and testsuite is updated -- I'll re-enable it again. 

I updated tests/simd.wasm with patches version of wabt.

alexcrichton (approved):

PR #209, opened 2020-05-05
Don't validate memories in `data.drop` instructions

alexcrichton:
For `data.drop` nowadays we only need to validate that the data segment
is a valid index, there's no need for the module to have a defined
memory.


yurydelendik:
Thank you for the patch

PR #210, opened 2020-05-07
Sync SIMD operators with the current spec

alexcrichton:
* Reorder the `read_0xfd_operator` in terms of increasing opcode
  numbers.
* Remove `I8x16Mul`
* Add `I{8x16,16x8,32x4}Abs`
* Remove `I64x2{All,Any}True`


yurydelendik (approved):
Do we want to remove I64x2TruncSatF64x2S, I64x2TruncSatF64x2U, F64x2ConvertI64x2S, F64x2ConvertI64x2U from Operator enum as well here? These were not listed in the encoded table and code did not fit there u8 range.

alexcrichton:
Oh oops meant to remove those but forgot! Done now.

PR #211, opened 2020-05-11
Rework how tests are run

alexcrichton:
This commit reorganizes how tests are run with a few goals in mind:

* Currently it's difficult to run just one test, you typically have to
  run everything and that can take some time.
* The error messages for failing tests aren't always immediately clear
  what test failed where.

This commit moves the `src/tests.rs` module to a `tests/all.rs`
integration test that is now run without a test harness. Test discovery
happens manually and rayon is used for parallelism. Test filters via CLI
arguments are supported and match on the filename of the test to run.

A few fixes to validation were applied along the way, along with
enabling a few more directories from the upstream spec tests. For now,
though, there's still a number of holes in the reference types
validation.


yurydelendik (approved):
Looks good.

(Note for release: API was changed to add `element_type_at` in the `WasmModuleResources`)

PR #212, opened 2020-05-12
Migrate most `*.wasm` tests to `*.wat`

alexcrichton:
This commit converts most binary `*.wasm` tests to `*.wat` instead. This
was programmatically done and only committed for tests where the `*.wat`
assembles back to the exact same binary as the original `*.wat`.


alexcrichton:
This is a pretty huge PR, but from what I can tell most of these modules also come from the `*.wast` tests, so would it perhaps be better to delete anything duplicated amongst `*.wast` tests as well?

yurydelendik:
The "xxxx.0.wasm" were (notice number of module from wast) originally created from the testsuite by wabt. Since we can do that use wast crate now automatically, the most of wasm files can be removed. I think we will end up with handful handcrafted wasms after that, e.g. reloc.wasm or naming.wasm.

alexcrichton:
Ok I've split this into a few commits. The first adds support for `*.wat` test cases. The second deletes all `*.wasm` files which are exactly byte-for-byte equivalent from extractions from `*.wast` files. The third blanket deletes all `*.NN.wasm` files assuming they were extracted at some point in the past (or extracted with `wast2json` vs `wat` which may have tiny differences in encoding). The last programmatically converts `*.wasm` to `*.wat` where possible and the `*.wat` encoding exactly matches the previous `*.wasm`.

yurydelendik:
Thank you for the patch

PR #213, opened 2020-05-12
Update implementation of reference types

alexcrichton:
Update the implementation given WebAssembly/reference-types#87 which has
a number of changes to the reference types implementation:

* Subtyping is removed
* `nullref` as a type was removed
* `anyref` was renamed to `externref`
* `ref.null` and `ref.is_null` take immediate type arguments as to what
  kind of `null` they're producing.

The spec testsuite repo has not yet updated with this change yet, so the
tests are temporarily vendored here in the `tests/wast` directory.


src/operators_validator.rs
```diff
@@ -1033,10 +1015,17 @@ impl OperatorValidator {
...
+                    Some(_) => {
+                        bail_op_err!("type mismatch: only integer types allowed with bare `select`")
+                    }
+                    None => {}
+                }
                 self.func_state.change_frame_after_select(ty)?;
             }
             Operator::TypedSelect { ty } => {
-                self.check_operands_3(Type::I32, ty, ty)?;
+                self.check_operands_3(ty, ty, Type::I32)?;
```
yurydelendik: <del>just noticed how wrong Operator::TypedSelect is for dead code, we need to match the `check_select` logic here (later?)</del>
nvm, it's fine since we don't have to infer type here

alexcrichton: Ah I went ahead and updated this so they use the same code path for consistency anyway.

yurydelendik (approved):
Thank you for the patch

PR #214, opened 2020-05-13
Fix `no_run` example pragma

fitzgen:
This fixes a compilation warning.


yurydelendik:
Thank you for the patch

PR #218, opened 2020-05-20
Flag repository as being moved

alexcrichton:
This commit flags this repository as being moved to the
https://github.com/bytecodealliance/wasm-tools umbrella repository. This
should allow us to develop text/binary/etc all in tandem and ensure that
they've all got a similar level of support for new wasm features.


alexcrichton:
Once this is merged I'll transfer all existing issues to the new repository and also comment on existing PRs

